PR #2, opened 2021-08-04
Add initial files needed to setup Read the Docs

allyhawkins:
This PR adds the initial files for starting the docs using Read the docs and begins the addition of some processing information. I started by creating the repo and following the [quick start instructions using Sphinx with read the docs](https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html#). This initiated the directory structure and filled out the initial files that are needed for read the docs. I also added in the MyST extension and followed the [Using Markdown with Sphinx instructions](https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html#using-markdown-with-sphinx) to enable use of markdown files. 

The only changes that were made to the files during the quickstart process were inserting the project name (scpca) and the author (The Childhood Cancer Data Lab).


jashapiro (commented):
I am assuming the contents of these files are fine based on the description of how they were created. But I do have a few questions:

Can we ignore the `_build` directory? If so, I think we should add that to `.gitignore` (Remove the files first, to avoid annoyance). This will make future changes much smaller and easier to track.

We probably will want to add a `requirements.txt` file and perhaps a `.readthedocs.yaml` configuration file, but I expect that can happen later and doesn't much matter for now.

allyhawkins:
> Can we ignore the _build directory? If so, I think we should add that to .gitignore (Remove the files first, to avoid annoyance). This will make future changes much smaller and easier to track.

Yes, I can add a gitignore file and put that in. I wasn't sure what to ignore since technically the `-build` directory is where the `index.html` lives which I assumed we needed? 

> We probably will want to add a requirements.txt file and perhaps a .readthedocs.yaml configuration file, but I expect that can happen later and doesn't much matter for now.

I also wasn't quite sure about these since I noticed they were in the `refinebio-docs` repo, but wasn't sure at what point we would need them. I can add those so we have them to start with? I didn't add any additional docs at this stage that aren't already pre-set during the quickstart but I can.

requirements.txt
```diff
@@ -0,0 +1 @@
+Sphinx==1.7.6
```
jashapiro, line 1: I assume we need `myst-parser` here too?

jashapiro (approved):
LGTM! Just the remaining question about whether we need `myst-parser` in the requirements. But otherwise seems like we should be set.

PR #11, opened 2021-08-20
Add spell check for markdown files 

### LINKED ISSUE #5: Add spell check to repo
Based on the comment below and the fact that it's just generally probably a good idea, we should set up a spell check github action for this repo. 

Not specific to this line but I think we should set up spell check for this soon because I'm finding myself asking "CellRanger or Cell Ranger or Cellranger?" and spell check would help with enforcement of that kind of thing. See: https://github.com/AlexsLemonade/refinebio-docs/issues/147

_Originally posted by @jaclyn-taroni in https://github.com/AlexsLemonade/scpca-docs/pull/3#discussion_r684408571_

allyhawkins:
Closes #5. This PR adds in the [check-spelling github action](https://github.com/check-spelling/check-spelling/wiki). To set this up, I followed the [quick start instructions](https://github.com/check-spelling/check-spelling/wiki/Quick-Start) and copied over the `.github` folder from the `check-spelling/spell-check-this` repo. I then looked at the instructions on the README.md file in the `actions` folder to modify the action so that it would ignore all files except for markdown files. This is done by ignoring all files in the `docs` folder in the `actions/excludes.txt` file and then allowing only markdown files in the `actions/only.txt` file. I also ignored any links within markdown files by using the `patterns.txt` file. I also added in any additional words that should be allowed into the dictionary by modifying the `allow.txt` file. 

I believe I have set this up correctly and so spell check should now be passing and be set up for the repo once this is merged, but that is dependent on my ability to set up perl regular expressions correctly which I am not the most confident in. It would be great if someone could double check that I have ignored and allowed the correct files with the correct regular expressions, particularly in the `only.txt` and `excludes.txt` (Note that I only added in the last two lines of the `excludes.txt` from the original file that can be found on the [check-spelling repo](https://github.com/check-spelling/spell-check-this/blob/main/.github/actions/spelling/excludes.txt).)


.github/workflows/spelling.yml
```diff
@@ -0,0 +1,31 @@
...
+      uses: actions/checkout@v2
+      with:
+        ref: refs/pull/${{github.event.pull_request.number}}/merge
+    - name: checkout
+      if: ${{ github.event_name == 'push' ||
+        (
+          contains(github.event.comment.body, '@check-spelling-bot apply')
+        ) }}
+      uses: actions/checkout@v2
+    - uses: check-spelling/check-spelling@main
```
jaclyn-taroni, line 24: This should be tied to a specific version rather than `main` -- probably `0.0.19` in this case. Quoting the quick start guide you linked:

> Best-practice for workflows is to use a pinned version.

.github/workflows/spelling.yml
```diff
@@ -0,0 +1,31 @@
+name: Spell checking
+on:
+  pull_request_target:
+  push:
+  issue_comment:
+    types: [created]
```
jaclyn-taroni, line 6: This seems like it's doing way too much? Our goal is to spell check the docs themselves when a pull request is filed.

jaclyn-taroni (commented):
I think we should think a little bit more about _when_ the action gets run. A disadvantage of this (vs. something like what we have set up in refine.bio examples) is that every time something is misspelled it will comment which potentially notifies multiple people, rather than just the PR author.

allyhawkins:
@jaclyn-taroni, based on our slack conversation and these comments, I went ahead and removed the use of the `check-spelling` github action and switched to the same custom spell check that we use in `training-modules` and `refinebio-examples`. In doing this, I modified the script to only spell check markdown files. I started a dictionary of words to ignore and also found a spelling error in `processing_information.md` that I fixed in the process of testing this.

docs/processing_information.md
```diff
@@ -21,7 +21,7 @@ We aligned reads to the transcriptome index using `salmon` with the default "sel
 Briefly, selective alignment uses a mapping score validated approach to identify maximal exact matches between reads and the provided index. 
 For all samples, we used selective alignment to the `splici` index. 
 
-A more detailed description of the alignment strategy invoked by `salmon` in conjuction with `alevin-fry` can be found in [Srivastava _et al._ 2020](https://doi.org/10.1186/s13059-020-02151-8).
+A more detailed description of the alignment strategy invoked by `salmon` in conjunction with `alevin-fry` can be found in [Srivastava _et al._ 2020](https://doi.org/10.1186/s13059-020-02151-8).
```
jaclyn-taroni, line 24: :its-working-anakin-meme:

scripts/spell-check.R
```diff
@@ -0,0 +1,32 @@
...
+library(magrittr)
+
+# Find .git root directory
+root_dir <- rprojroot::find_root(rprojroot::has_dir(".git"))
+
+# Read in dictionary
+dictionary <- readLines(file.path(root_dir, 'components', 'dictionary.txt'))
+
+# Add emoji to dictionary
+dictionary <- c(dictionary, spelling::spell_check_text("⚠️")$word)
```
jaclyn-taroni, line 16: I think this can come out?

jaclyn-taroni (approved):
Looks good! Can you add some info about spell check to the main README before you merge?

Issue #6, opened 2021-08-19
Section: Download files


dvenprasad:
Talk about what user can expect in their download files and how it is structured.

Probably also needs illustration of the file structure.

allyhawkins:
Based on our conversation in the scpca standing meeting, we may also want to include some basic R and python code on how to read in these files in this section.

dvenprasad:
![download-folder](https://user-images.githubusercontent.com/15315514/142041878-cfc0a3e7-86dd-46a9-9835-8cec35fe8e32.png)

dvenprasad:
Ignore the previous one. It is incorrect. 

**Project Download Folder**
![project-download-folder](https://user-images.githubusercontent.com/15315514/142902842-6dcea60a-e7d8-4dfa-8de1-aebb581109c8.png)


**Sample Download Folder**
![sample-download-folder](https://user-images.githubusercontent.com/15315514/142902843-fcdabcf0-ea7b-4e05-b436-e8c93b880560.png)

jaclyn-taroni:
Can this get closed?

PR #20, opened 2021-10-07
Add filtering information to processing docs 

### LINKED ISSUE #18: Processing Information for filtering post alevin-fry
I noticed that we have the section complete on processing with Alevin-fry, but we don't yet have information on generating the filtered single cell experiment objects, i.e. filtering with emptyDrops. We should fill in a section for how we obtained the filtered output and what parameters we used that are different from the default for emptyDrops (i.e. lower=200).

allyhawkins:
Closes #18. In this PR, I'm adding in the section about how we created the filtered counts matrix post processing with Alevin-fry. I added in a brief paragraph about using `DropletUtils::emptyDrops()` including a sentence about what it does, our choice of `lower=200`, and then how we actually filtered on the FDR <= 0.01. Please let me know if the explanation is at the right level of detail or not.

Another thing I wanted to note here is that we haven't yet mentioned the actual file types that we are working with or that we have both an unfiltered and filtered output. I was wondering if we should include that here somewhere or have a separate section describing what files are included for each sample (Maybe this is a good candidate for a FAQ)? 

Also in this PR I removed the headers for bulk RNA seq and spatial since that will not be part of MVP I thought it made sense to remove for now and then we can add it in when we set up the workflow for those.

### DIFF
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -43,7 +43,7 @@ For single-nuclei samples, all counts for spliced cDNA and intronic regions were
 
 #### Filtering cells
 
-In addition to an unfiltered counts matrix, we provide a matrix filtered to only droplets that are likely to include cells, removing any cell barcodes that are likely to correspond to empty droplets.
+In addition to an unfiltered counts matrix, we provide a matrix filtered to only cell barcodes from droplets that are likely to include true cells.
 To do this we used [`DropletUtils::emptyDrops()`](https://www.bioconductor.org/packages/devel/bioc/vignettes/DropletUtils/inst/doc/DropletUtils.html#detecting-empty-droplets), a function that estimates the profile of cells containing ambient RNA and tests the likelihood of all other droplets as differing from the ambient profile [Lun _et al._ 2019](https://doi.org/10.1186/s13059-019-1662-y). 
 We defined the ambient profile to include all droplets with less than 200 UMI per cell by using the option `lower=200`.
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
> Another thing I wanted to note here is that we haven't yet mentioned the actual file types that we are working with or that we have both an unfiltered and filtered output. I was wondering if we should include that here somewhere or have a separate section describing what files are included for each sample (Maybe this is a good candidate for a FAQ)?

This is tracked in #6 so if you have thoughts after writing this, you can jot them down there!

allyhawkins:
@jashapiro I changed the intro section to combine what I had with what you had suggested. Let me know what you think!

docs/processing_information.md
```diff
@@ -43,7 +43,7 @@ For single-nuclei samples, all counts for spliced cDNA and intronic regions were
 
 #### Filtering cells
 
-We then removed any cell barcodes that are likely to correspond to empty droplets, rather than true cells. 
+In addition to an unfiltered counts matrix, we provide a matrix filtered to only droplets that are likely to include cells, removing any cell barcodes that are likely to correspond to empty droplets.
```
jashapiro, line 46: The second clause seemed redundant to me, but including "cell barcodes" seems important, so what do you think of this?
```suggestion
In addition to an unfiltered counts matrix, we provide a matrix filtered to only cell barcodes from droplets that are likely to include true cells.
```

allyhawkins: I like it! Will go ahead and commit that change.

PR #25, opened 2021-10-13
theme docs

davidsmejia:
* Adds instructions for setting up to the readme
* Adds and customizes sphinx_rtd_theme to project
* Adds dev server to see live updates after saving file


docs/index.md
```diff
@@ -0,0 +1,11 @@
+# ScPCA Portal Documentation
+
+This is the documentation for the [Single-cell Pediatric Cancer Atlas Portal](https://scpca.alexslemonade.org).
+The ScPCA Portal is a database of single-cell data from pediatric cancer clinical samples and xenografts.
```
allyhawkins, line 4:
```suggestion
The ScPCA Portal is a database of single-cell and single-nuclei RNA-sequencing data from pediatric cancer clinical samples and xenografts.
```
Since technically this has both single-cell and single-nuclei data and not just single-cell we should probably include that here. We should also include the type of data.

davidsmejia: This would be a good consideration for the homepage of the portal as well

PR #29, opened 2021-10-15
Describe what genes are included in the reference transcriptome

### LINKED ISSUE #12: Add link to code that generates splici index
Just in case people are more curious on the exact details of how we built our splici index, we should include a link to the code to build it from the ensembl gtf (or maybe a notebook with instructions on how to build it?). 

@jaclyn-taroni **edit** - breaking out strikethrough text into own issue:
~When we do that, or add any links in the docs, we will want to implement some sort of URL checker which ensures that the links we provide aren't broken and that anything that's happening upstream of the links is working properly.~

### LINKED ISSUE #23: FAQ: What genes are used
We are assaying 60K+ "genes", but many of these are pseudogenes, so we should probably make a note of that somehow. This is related to https://github.com/AlexsLemonade/scpca-docs/issues/12, and I would suggest we tackle both at the same time.

allyhawkins:
Here I am adding in a FAQ addressing what type of genes are included in the index, #23. While doing this, I also included a link to the code used to generate the reference transcriptome used to make the splici index (addressing #12). I kept this pretty short and mentioned the number of genes and that they include protein-coding genes, pseudogenes, and non-coding genes. I also put in a brief sentence that the output files they receive will include all genes regardless of if they are detected or not. 

One thing I was wondering was if we want to add in a statement that we suggest removing any genes that are not detected or have mean gene expression of 0 before doing analysis or do we just leave that alone and not include anything on suggestions yet? 

In general I didn't go into too much detail and kept this brief, but let me know if there was more detail that you think I should be providing. 

Closes #23 
Closes #12

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -22,7 +22,7 @@ Recent reports from others support our findings.
 
 #### What genes are included in the reference transcriptome? 
 
-The {ref}`reference transcriptome index <processing_information:reference transcriptome index>` that was used for alignment included 60,319 genes.
-The reference transcriptome was constructed by extracting both spliced cDNA and intronic regions from the primary genome assembly ([see the code used to generate the reference transcriptome](https://github.com/AlexsLemonade/scpca-nf/blob/5dc49ffdcdb5c74dca86b0b79878f4d060029a53/bin/make_splici_fasta.R)).
-In addition to protein-coding genes, this reference transcriptome index includes pseudogenes and non-coding RNA. 
-Although there are 60,319 genes that are present in the reference transcriptome index, it is likely that many of these genes will not be detected in any given library.
+The {ref}`reference transcriptome index <processing_information:reference transcriptome index>` that was used for alignment was constructed by extracting both spliced cDNA and intronic regions from the primary genome assembly ([see the code used to generate the reference transcriptome](https://github.com/AlexsLemonade/scpca-nf/blob/5dc49ffdcdb5c74dca86b0b79878f4d060029a53/bin/make_splici_fasta.R)).
+The resulting reference transcriptome index included 60,319 genes.
+In addition to protein-coding genes, this list of genes includes pseudogenes and non-coding RNA.
+The gene expression data files available for download report all possible genes present in the reference transcriptome, even if not detected in a given library. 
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -18,4 +18,11 @@ We also compared the mean gene expression reported for each gene by both methods
...
-[You _et al._ (2021)](https://doi.org/10.1101/2021.06.17.448895) and [Tian _et al._ (2019)](https://doi.org/10.1038/s41592-019-0425-8) have also noted that results from different pre-processing workflows for single-cell RNA-sequencing analysis tend to result in compatible results downstream.
\ No newline at end of file
+[You _et al._ (2021)](https://doi.org/10.1101/2021.06.17.448895) and [Tian _et al._ (2019)](https://doi.org/10.1038/s41592-019-0425-8) have also noted that results from different pre-processing workflows for single-cell RNA-sequencing analysis tend to result in compatible results downstream.
+
+#### What genes are included in the reference transcriptome? 
+
+The {ref}`reference transcriptome index <processing_information:reference transcriptome index>` that was used for alignment was constructed by extracting both spliced cDNA and intronic regions from the primary genome assembly ([see the code used to generate the reference transcriptome](https://github.com/AlexsLemonade/scpca-nf/blob/5dc49ffdcdb5c74dca86b0b79878f4d060029a53/bin/make_splici_fasta.R)).
```
jashapiro, line 25: Using the tag we used for processing to make it easier to update this in the future if needed. 
```suggestion
The {ref}`reference transcriptome index <processing_information:reference transcriptome index>` that was used for alignment was constructed by extracting both spliced cDNA and intronic regions from the primary genome assembly ([see the code used to generate the reference transcriptome](https://github.com/AlexsLemonade/scpca-nf/blob/v0.1.1/bin/make_splici_fasta.R)).
```

docs/faq.md
```diff
@@ -18,4 +18,11 @@ We also compared the mean gene expression reported for each gene by both methods
...
+The resulting reference transcriptome index included 60,319 genes.
```
jashapiro, line 26: Different wording for spice.
```suggestion
The resulting reference transcriptome index contains 60,319 genes.
```

docs/faq.md
jashapiro, line 25: We should probably say that we are use GRCh38, ensembl database version 104. I was going to leave that out, but if we are going to give the number of genes, below, we should probably name the version here as well.

PR #30, opened 2021-10-15
Add link to code for QC report 

### LINKED ISSUE #19: FAQ: Where can I see the code for the QC reports?
Had this thought bubble up this morning. But we could make it more general: Where can I see the code for processing and QC? And then the answer would break it down by repository, I'd think.

Going to stick it in the current Epic for now, but we may decide to punt.

allyhawkins:
Closes #19. Here I'm adding a FAQ that includes information about how we generated the QC report. I included a link to the template for the QC report and then also provided some explanation on how we got the input needed for the QC report. I think some of it might be repetitive, but I thought it would be nice to have some context of what goes into the QC report and not just give the link to the code, but open to other suggestions on how to frame this. 

I also included a link to the `scpcaTools` package as I thought it could be useful for people to have somewhere.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -37,7 +37,7 @@ scpca_sample <- readRDS("SCPCL000000_filtered.rds")
 
 ## What genes are included in the reference transcriptome? 
 
-The {ref}`reference transcriptome index <processing_information:reference transcriptome index>` that was used for alignment was constructed by extracting both spliced cDNA and intronic regions from the primary genome assembly GRCh38, Ensembl database version 104 ([see the code used to generate the reference transcriptome](https://github.com/AlexsLemonade/scpca-nf/blob/5dc49ffdcdb5c74dca86b0b79878f4d060029a53/bin/make_splici_fasta.R)).
+The {ref}`reference transcriptome index <processing_information:reference transcriptome index>` that was used for alignment was constructed by extracting both spliced cDNA and intronic regions from the primary genome assembly GRCh38, Ensembl database version 104 ([see the code used to generate the reference transcriptome](https://github.com/AlexsLemonade/scpca-nf/blob/main/bin/make_splici_fasta.R)).
 The resulting reference transcriptome index contains 60,319 genes.
 In addition to protein-coding genes, this list of genes includes pseudogenes and non-coding RNA.
 The gene expression data files available for download report all possible genes present in the reference transcriptome, even if not detected in a given library. 


docs/faq.md
```diff
@@ -41,3 +41,10 @@ The {ref}`reference transcriptome index <processing_information:reference transc
 The resulting reference transcriptome index contains 60,319 genes.
 In addition to protein-coding genes, this list of genes includes pseudogenes and non-coding RNA.
 The gene expression data files available for download report all possible genes present in the reference transcriptome, even if not detected in a given library. 
+
+## Where can I see the code for generating QC reports? 
+
+A QC report for every processed library is included with all downloads. 
+Following alignment and quantification, the unfiltered output from Alevin-fry was imported into R as a `SingleCellExperiment` object and filtered using `DropletUtils::emptyDrops()`. 
+Next, we generated the QC report, using the unfiltered and filtered `SingleCellExperiment` objects as input. 
+You can find the [code used to generate the QC report](https://github.com/AlexsLemonade/scpcaTools/blob/fd536edebddb2589c430c460ae7c79458dae803c/inst/rmd/qc_report.rmd) in the package we developed for working with processed ScPCA data, `scpcaTools`(https://github.com/AlexsLemonade/scpcaTools/tree/v0.1.1).
```
jashapiro, line 50: I'm thinking maybe we don't want to link to specific versions here (or elsewhere in the docs). I worry more about linking to old versions and forgetting to update than people getting a version that is not quite in sync with something that they downloaded. 

I also thought that we might just link to the function, since that is how most people would use them, if they so desired. I then linked to the directory with the templates, but I don't feel strongly about that: if we want to link to the main template as you had before, that would be fine.   
 
```suggestion
You can find the [function for generating a QC report](https://github.com/AlexsLemonade/scpcaTools/blob/main/R/generate_qc_report.R) and the [QC report template documents](https://github.com/AlexsLemonade/scpcaTools/tree/main/inst/rmd) in the package we developed for working with processed ScPCA data, `scpcaTools`(https://github.com/AlexsLemonade/scpcaTools). 
```

docs/faq.md
```diff
@@ -41,3 +41,10 @@ The {ref}`reference transcriptome index <processing_information:reference transc
 The resulting reference transcriptome index contains 60,319 genes.
 In addition to protein-coding genes, this list of genes includes pseudogenes and non-coding RNA.
 The gene expression data files available for download report all possible genes present in the reference transcriptome, even if not detected in a given library. 
+
+## Where can I see the code for generating QC reports? 
+
+A QC report for every processed library is included with all downloads. 
+Following alignment and quantification, the unfiltered output from Alevin-fry was imported into R as a `SingleCellExperiment` object and filtered using `DropletUtils::emptyDrops()`. 
+Next, we generated the QC report, using the unfiltered and filtered `SingleCellExperiment` objects as input.
```
jashapiro, line 49: I'm making this a bit less detailed, but adding links to the relevant other parts of the docs (hopefully correctly). Trying to reduce redundancy, and get people to the main answer more quickly.

```suggestion
A QC report for every processed library is included with all downloads, generated from the unfiltered and {ref}`filtered <processing_information:filtering cells>` {ref}`gene expression files <gene_expression_file_contents:gene expression file contents>.
```

jashapiro:
>  I wasn't sure about linking to a tagged version or not, but did so because we did it for the splici code from the `scpca-nf` repo. Should we also change that link then?

I think that is where I was leaning, which was what I meant when I made a reference to previous FAQ questions as well. I think I would rather people see the lastest code from the docs, even if it is not exactly the code that was used for a particular sample. If people need to track that down, they will have to look at the version tagged in the metadata (as the docs might not correspond anyway as we update them).

PR #32, opened 2021-10-15
FAQ for difference between sample and library ID

### LINKED ISSUE #28: FAQ: What is the difference between a sample ID and library ID?
The concept of having both a sample ID and library ID might be confusing to users, so we should probably include some text explaining the difference between them.

allyhawkins:
Closes #28. I added in the text for the FAQ on explaining the difference between samples and libraries in the portal. While doing this, I thought it might be helpful to also include participant ID's in the discussion as I was having trouble explaining what a sample was without talking about participants. 

I'm not quite confident on the explanation of a library here for the audience of the portal but I'm assuming that most people would be generally familiar with the single-cell workflow. So I chose to refer to the different libraries as separate emulsions and droplet generations, which made the most sense to me. Definitely open to other suggestions on how best to explain this! 

Because I changed the header of this FAQ, I also had to update the link in the download section. This PR is also stacked on #30, since I was adding more to the same FAQ section and wanted to make conflict resolution down the line easier, so that will need to be reviewed first.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -40,7 +40,7 @@ A sample ID, labeled as `scpca_sample_id` and indicated by the prefix `SCPCS`, r
 For example, one participant may have a sample collected both at initial diagnosis and at relapse.
 This would result in two unique sample ID's with the same participant ID. 
 
-The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single emulsion and droplet generation using the 10X Genomics worfklow from a given sample. 
+The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single emulsion and droplet generation using the 10X Genomics workflow from a given sample. 
 In most cases, each sample will only have one corresponding library.
 However, in some cases multiple libraries underwent separate droplet generation and sequencing from the same sample, resulting in more than one library ID being associated with the same sample ID. 
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -33,7 +33,16 @@ library(SingleCellExperiment)
 scpca_sample <- readRDS("SCPCL000000_filtered.rds")
 ```
 
-## What is the difference between a sample and a library?
+## What is the difference between participants, samples, and libraries?
+
+The `participant_id` indicates the participant, or patient, from which a sample, or collection of samples, is obtained from. 
+A sample ID, labeled as `scpca_sample_id` and indicated by the prefix `SCPCS`, represents a unique tissue that was collected from a participant. 
+For example, one participant can have a sample collected both at initial diagnosis and at relapse.
+This would result in two unique sample ID's with the same participant ID. 
+
+The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single emulsion and droplet generation using the 10X worfklow from a given sample. 
+In most cases, each sample will only have one corresponding library.
+However, in some cases multiple libraries underwent separate droplet generation and sequencing from the same sample, resulting in multiple library ID's being associated with the same sample ID. 
 
 ## What genes are included in the reference transcriptome? 
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -37,12 +37,12 @@ scpca_sample <- readRDS("SCPCL000000_filtered.rds")
 
 The `participant_id` indicates the participant, or patient, from which a sample, or collection of samples, is obtained from. 
 A sample ID, labeled as `scpca_sample_id` and indicated by the prefix `SCPCS`, represents a unique tissue that was collected from a participant. 
-For example, one participant can have a sample collected both at initial diagnosis and at relapse.
+For example, one participant may have a sample collected both at initial diagnosis and at relapse.
 This would result in two unique sample ID's with the same participant ID. 
 
-The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single emulsion and droplet generation using the 10X worfklow from a given sample. 
+The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single emulsion and droplet generation using the 10X Genomics worfklow from a given sample. 
 In most cases, each sample will only have one corresponding library.
-However, in some cases multiple libraries underwent separate droplet generation and sequencing from the same sample, resulting in multiple library ID's being associated with the same sample ID. 
+However, in some cases multiple libraries underwent separate droplet generation and sequencing from the same sample, resulting in more than one library ID being associated with the same sample ID. 
 
 ## What genes are included in the reference transcriptome? 
 
--- a/docs/download_files.md
+++ b/docs/download_files.md
@@ -12,7 +12,7 @@ _add in illustration of file structure for sample downloads_
 
 Sample folders (indicated by the `SCPCS` prefix) contain the files for all libraries (`SCPCL` prefix) derived from that biological sample. 
 Most samples only have one library that has been sequenced. 
-See the {ref}`FAQ section about samples and libraries <faq:what is the difference between a sample and a library?>` for more information.
+See the {ref}`FAQ section about samples and libraries <faq:What is the difference between participants, samples, and libraries?>` for more information.
 
 The files associated with each library are (example shown for a library with ID `SCPCL000000`):
 - An unfiltered counts file: `SCPCL000000_unfiltered.rds`, 


docs/faq.md
```diff
@@ -33,7 +33,16 @@ library(SingleCellExperiment)
 scpca_sample <- readRDS("SCPCL000000_filtered.rds")
 ```
 
-## What is the difference between a sample and a library?
+## What is the difference between participants, samples, and libraries?
+
+The `participant_id` indicates the participant, or patient, from which a sample, or collection of samples, is obtained from.
```
jashapiro, line 38: We don't have participant ids for all samples right? Only when there is more than one sample from the same participant? If so, we should say that explicitly.  (Maybe in a separate FAQ entry)

I simplified here, mostly because there were lots of commas that I think were getting confusing.
```suggestion
The `participant_id` indicates the participant from which a sample or collection of samples was obtained. 
```

docs/faq.md
```diff
@@ -33,7 +33,16 @@ library(SingleCellExperiment)
 scpca_sample <- readRDS("SCPCL000000_filtered.rds")
 ```
 
-## What is the difference between a sample and a library?
+## What is the difference between participants, samples, and libraries?
+
+The `participant_id` indicates the participant, or patient, from which a sample, or collection of samples, is obtained from. 
+A sample ID, labeled as `scpca_sample_id` and indicated by the prefix `SCPCS`, represents a unique tissue that was collected from a participant. 
+For example, one participant may have a sample collected both at initial diagnosis and at relapse.
+This would result in two unique sample ID's with the same participant ID.
```
jashapiro, line 41: Only because you just said "unique"
```suggestion
This would result in two different sample ID's with the same participant ID. 
```

docs/faq.md
```diff
@@ -33,7 +33,16 @@ library(SingleCellExperiment)
...
+## What is the difference between participants, samples, and libraries?
+
+The `participant_id` indicates the participant, or patient, from which a sample, or collection of samples, is obtained from. 
+A sample ID, labeled as `scpca_sample_id` and indicated by the prefix `SCPCS`, represents a unique tissue that was collected from a participant. 
+For example, one participant may have a sample collected both at initial diagnosis and at relapse.
+This would result in two unique sample ID's with the same participant ID. 
+
+The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single emulsion and droplet generation using the 10X Genomics workflow from a given sample. 
+In most cases, each sample will only have one corresponding library.
+However, in some cases multiple libraries underwent separate droplet generation and sequencing from the same sample, resulting in more than one library ID being associated with the same sample ID.
```
jashapiro, line 45: Adding a potential "why" here.
```suggestion
However, in some cases multiple libraries were generated from the same sample (perhaps with different experimental protocols), resulting in more than one library ID being associated with the same sample ID. 
```

docs/faq.md
```diff
@@ -33,7 +33,16 @@ library(SingleCellExperiment)
...
-## What is the difference between a sample and a library?
+## What is the difference between participants, samples, and libraries?
+
+The `participant_id` indicates the participant, or patient, from which a sample, or collection of samples, is obtained from. 
+A sample ID, labeled as `scpca_sample_id` and indicated by the prefix `SCPCS`, represents a unique tissue that was collected from a participant. 
+For example, one participant may have a sample collected both at initial diagnosis and at relapse.
+This would result in two unique sample ID's with the same participant ID. 
+
+The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single emulsion and droplet generation using the 10X Genomics workflow from a given sample. 
+In most cases, each sample will only have one corresponding library.
```
jashapiro, line 44: Realizing that as we get bulk, this may no longer be true.
```suggestion
In most cases, each sample will only have one corresponding single-cell library, and may also have an associated bulk RNA-seq library.
```

docs/faq.md
```diff
@@ -33,7 +33,16 @@ library(SingleCellExperiment)
...
 
-## What is the difference between a sample and a library?
+## What is the difference between participants, samples, and libraries?
+
+The `participant_id` indicates the participant, or patient, from which a sample, or collection of samples, is obtained from. 
+A sample ID, labeled as `scpca_sample_id` and indicated by the prefix `SCPCS`, represents a unique tissue that was collected from a participant. 
+For example, one participant may have a sample collected both at initial diagnosis and at relapse.
+This would result in two unique sample ID's with the same participant ID. 
+
+The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single emulsion and droplet generation using the 10X Genomics workflow from a given sample.
```
jashapiro, line 43: I think we want to make this a bit less specific, as we use library ids also for bulk, so we should to say something about that. I'm fighting with it a bit here, but trying to get across that the CITE-seq sequencing library falls in the same library, but bulk does not. I don't know it "set of cells" is a good metric, but it is what I came up with...


```suggestion
The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single set of cells from a tissue sample.
For a bulk RNA-seq experiment, this will result in a single sequencing library. 
For single-cell or single-nuclei experiments, this will be the result of emulsion and droplet generation using the 10X Genomics workflow, potentially including both RNA-seq, CITE-seq and cell hashing sequencing libraries. 
```

jashapiro (commented):
This looks good, but I think there are a few things we might still want to clarify. The two main things I think we want to specifically address are that bulk is a separate library, but CITE-seq & cellhash are the same as the SCRNA library. I tried to say something in a suggestion about the "set of cells" being the unit for a library, which is true (and kind of how I think about it), but may be confusing rather than illuminating. I think it is probably better to be explicit where we can be though, so saying something about CITE-seq makes sense to me. For the reverse, maybe we should also say that if a sample had both single cell and single nuclei, or two different 10X protocols, that would be two libraries? 

Maybe instead of "emulsion and droplet generation" as the definition we should just say something about generic "library generation protocol", but make emulsion and droplet generation an example? 

(Also confusing that the library may actually have more than one sample in the case of cell hashing, but we can let that slide for now)

I also think we might still want to break out the participant ID discussion to a separate FAQ question. Something to address the question  "Why do some samples have participant IDs and others don't?"

docs/faq.md
```diff
@@ -33,7 +33,24 @@ library(SingleCellExperiment)
...
+In most cases, each sample will only have one corresponding single-cell library, and may also have an associated bulk RNA-seq library.
+However, in some cases multiple libraries were created by separate droplet generation and sequencing from the same sample, resulting in more than one library ID being associated with the same sample ID. 
+
+## Why do some samples have missing participant IDs?
+
+The `participant_id` indicates the participant from which a collection of samples was obtained. 
+For example, one participant may have a sample collected both at initial diagnosis and at relapse.
+This would result in two different sample ID's, but the same participant ID. 
+For the most part, each participant only has one corresponding sample that was collected and submitted for sequencing. 
+Because of this, many of the samples do not have a participant ID and only those samples in which multiple unique samples have been obtained from the same participant will a participant ID be present.
```
jashapiro, line 53: Separating the second clause and making it more active.
```suggestion
Because of this, many of the samples do not have a separate participant ID. 
Participant IDs are only present for samples that were derived from the same participant as at least one other sample. 
```

docs/faq.md
```diff
@@ -33,7 +33,24 @@ library(SingleCellExperiment)
...
+The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single set of cells from a tissue sample.
+For single-cell or single-nuclei experiments, this will be the result of emulsion and droplet generation using the 10X Genomics workflow, potentially including both RNA-seq, CITE-seq and cell hashing sequencing libraries. 
+For a bulk RNA-seq experiment, this will result in a single sequencing library. 
+
+In most cases, each sample will only have one corresponding single-cell library, and may also have an associated bulk RNA-seq library.
+However, in some cases multiple libraries were created by separate droplet generation and sequencing from the same sample, resulting in more than one library ID being associated with the same sample ID. 
+
+## Why do some samples have missing participant IDs?
+
+The `participant_id` indicates the participant from which a collection of samples was obtained.
```
jashapiro, line 49: Emphasizing again at the start that this is a conditional field.
```suggestion
The `participant_id`, when present, indicates the participant from which a collection of samples was obtained. 
```

docs/faq.md
```diff
@@ -33,7 +33,24 @@ library(SingleCellExperiment)
...
+
+In most cases, each sample will only have one corresponding single-cell library, and may also have an associated bulk RNA-seq library.
+However, in some cases multiple libraries were created by separate droplet generation and sequencing from the same sample, resulting in more than one library ID being associated with the same sample ID. 
+
+## Why do some samples have missing participant IDs?
+
+The `participant_id` indicates the participant from which a collection of samples was obtained. 
+For example, one participant may have a sample collected both at initial diagnosis and at relapse.
+This would result in two different sample ID's, but the same participant ID. 
+For the most part, each participant only has one corresponding sample that was collected and submitted for sequencing.
```
jashapiro, line 52:
```suggestion
However, for most participants, only a single sample was collected and submitted for sequencing. 
```

docs/faq.md
```diff
@@ -33,7 +33,24 @@ library(SingleCellExperiment)
...
+## What is the difference between samples and libraries?
+
+A sample ID, labeled as `scpca_sample_id` and indicated by the prefix `SCPCS`, represents a unique tissue that was collected from a participant. 
+
+The library ID, labeled as `scpca_library_id` and indicated by the prefix `SCPCL`, represents a single set of cells from a tissue sample.
+For single-cell or single-nuclei experiments, this will be the result of emulsion and droplet generation using the 10X Genomics workflow, potentially including both RNA-seq, CITE-seq and cell hashing sequencing libraries. 
+For a bulk RNA-seq experiment, this will result in a single sequencing library. 
+
+In most cases, each sample will only have one corresponding single-cell library, and may also have an associated bulk RNA-seq library.
+However, in some cases multiple libraries were created by separate droplet generation and sequencing from the same sample, resulting in more than one library ID being associated with the same sample ID.
```
jashapiro, line 45:
```suggestion
However, in some cases multiple libraries were created by separate droplet generation and sequencing from the same sample, resulting in more than one single-cell library ID being associated with the same sample ID. 
```

jashapiro (approved):
LGTM!

PR #33, opened 2021-10-18
Add FAQ on how to convert the SCE objects to Seurat objects

### LINKED ISSUE #26: FAQ: How do convert the SingleCellExperiments to a Seurat object?
In case people want to use Seurat, we should include a FAQ with instructions on how to convert the object from bioconductor's SingleCellExperiment object to Seurat. I think we could follow what's in the [vignette from Seurat](https://satijalab.org/seurat/archive/v3.1/conversion_vignette.html) on conversion. 

Where things might get tricky is when we want to convert objects that have an `AltExp`. To do this, we probably want to pull out the `AltExp` and then build the Seurat objects. We can use the [vignette on setting up Seurat with multimodal data](https://satijalab.org/seurat/articles/multimodal_vignette.html#setup-a-seurat-object-add-the-rna-and-protein-data) as a guide for this.

allyhawkins:
Closes #26. Here I added in a section to address how to convert the SCE objects that we provide to Seurat objects. Right now I have only included how to work with the objects that contain RNA-seq data with/without CITE-seq data, but I'm anticipating that we would have to update these instructions for cell hashing data at a later date.

In order to do the conversion, I originally was going to use `as.Seurat()` which is suggested in the [vignette from Seurat](https://satijalab.org/seurat/archive/v3.1/conversion_vignette.html), however this requires normalized counts to be stored as `logcounts` in the sce object. Since we do not provide normalized sce objects, I tried a few ways around this, but then after some searching around realized that the best way to do this would probably be to grab the pieces from the sce and then build the Seurat object using `CreateSeuratObject()`. In doing this, I'm adding the `colData` to the `meta.data` of Seurat within the function, and then separately I needed to add in the `rowData` to the `meta.features` of the RNA object and the `metadata(sce)` to the `misc` of the Seurat object. 

I then included a section to add in the CITE-seq data. Here, I followed the [vignette on multi-modal data](https://satijalab.org/seurat/articles/multimodal_vignette.html#setup-a-seurat-object-add-the-rna-and-protein-data) to first create an additional assay, adding in the `rowData` for the CITE-seq separately to the `meta.features` slot first. Then I am adding that assay to the seurat object. 

One question I have for reviewers is that here I'm starting after reading in the RDS file, since we already have a FAQ to address that. Should I instead include the line to read in the RDS file? Or is it okay to start with the sce object already in the environment. 

Also noting that this PR is stacked on #32.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -81,15 +81,8 @@ library(SingleCellExperiment)
 # read in RDS file 
 sce <- readRDS("SCPCL000000_filtered.rds")
 
-# grab counts matrix from SingleCellExperiment needed to build the Seurat Object
-counts <- counts(sce)
-
-# colData and rowData will need to be converted to a data.frame to be used in the Seurat object
-cell_metadata <- as.data.frame(colData(sce))
-row_metadata <- as.data.frame(rowData(sce))
-
-# create seurat object 
-seurat_object <- CreateSeuratObject(counts = counts, 
+# create seurat object from the SCE counts matrix
+seurat_object <- CreateSeuratObject(counts = counts(sce), 
                                     assay = "RNA",
                                     project = "SCPCL000000")
 ```
@@ -97,7 +90,7 @@ The above code will only maintain information found in the original counts matri
 Optionally, if you would like to keep the included cell and gene associated metadata during conversion to the Seurat object you can perform the below additional steps: 
 
 ```
-# colData and rowData will need to be converted to a data.frame to be used in the Seurat object
+# convert colData and rowData to data.frame for use in the Seurat object
 cell_metadata <- as.data.frame(colData(sce))
 row_metadata <- as.data.frame(rowData(sce))
 
@@ -111,19 +104,15 @@ seurat_object[["RNA"]]@meta.features <- row_metadata
 seurat_object@misc <- metadata(sce)
 ```
 
-In order to convert `SingleCellExperiment` objects from libraries that have counts from both RNA-sequencing and CITE-seq, you can use the following additional commands.
-This adds a second assay containing the CITE-seq counts and associated feature data to an existing Seurat object:
+For `SingleCellExperiment` objects from libraries with both RNA-seq and CITE-seq data, you can use the following additional commands to add a second assay containing the CITE-seq counts and associated feature data:
 
 ```
 # create assay object in Seurat from CITE-seq counts found in altExp(SingleCellExperiment)
-cite_counts <- counts(altExp(sce))
-cite_assay <- CreateAssayObject(counts = cite_counts)
-
-# optional: convert rowData to data.frame for use in Seurat object
-cite_row_metadata <- as.data.frame(rowData(altExp(sce)))
+cite_assay <- CreateAssayObject(counts = counts(altExp(sce)))
 
 # optional: add row metadata (rowData) from altExp to assay 
-cite_assay@meta.features = cite_row_metadata
+cite_row_metadata <- as.data.frame(rowData(altExp(sce)))
+cite_assay@meta.features <- cite_row_metadata
 
 # add altExp from SingleCellExperiment as second assay to Seurat
 seurat_object[["CITE"]] <- cite_assay
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -57,3 +57,45 @@ A QC report for every processed library is included with all downloads.
 Following alignment and quantification, the unfiltered output from Alevin-fry was imported into R as a `SingleCellExperiment` object and filtered using `DropletUtils::emptyDrops()`. 
 Next, we generated the QC report, using the unfiltered and filtered `SingleCellExperiment` objects as input. 
 You can find the [code used to generate the QC report](https://github.com/AlexsLemonade/scpcaTools/blob/fd536edebddb2589c430c460ae7c79458dae803c/inst/rmd/qc_report.rmd) in the package we developed for working with processed ScPCA data, `scpcaTools`(https://github.com/AlexsLemonade/scpcaTools/tree/v0.1.1). 
+
+## What if I want to use Seurat instead of Bioconductor? 
+
+The files available for download contain [`SingleCellExperiment objects`](http://bioconductor.org/books/3.13/OSCA.intro/the-singlecellexperiment-class.html). 
+These can easily be converted into Seurat objects. 
+
+You will need to [install and load the `Seurat` package](https://satijalab.org/seurat/articles/install.html) to work with Seurat objects.
+
+For libraries that only contain RNA-sequencing data (i.e. do not have a CITE-seq library found in the `altExp` of the `SingleCellExperiment object`), you can use the following commands:
+
+```
+library(Seurat)
+
+# grab counts matrix from the SingleCellExperiment
+counts <- as.matrix(counts(sce))
+
+# create seurat object 
+seurat_object <- CreateSeuratObject(counts = counts, assay = "RNA",
+                                    project = "project_name",
+                                    # add colData from SingleCellExperiment
+                                    meta.data = as.data.frame(colData(sce)))
+
+# add rowData from SingleCellExperiment to Seurat 
+seurat_object[["RNA"]]@meta.features <- as.data.frame(rowData(sce))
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -57,4 +57,47 @@ The gene expression data files available for download report all possible genes
...
+For libraries that only contain RNA-sequencing data (i.e. do not have a CITE-seq library found in the `altExp` of the `SingleCellExperiment` object), you can use the following commands:
+
+```
+library(Seurat)
+
+# grab counts matrix from the SingleCellExperiment
+counts <- as.matrix(counts(sce))
+
+# create seurat object 
+seurat_object <- CreateSeuratObject(counts = counts, assay = "RNA",
```
jashapiro, line 78: If breaking onto multiple lines, I prefer all arguments on separate lines.
```suggestion
seurat_object <- CreateSeuratObject(counts = counts, 
                                    assay = "RNA",
```

docs/faq.md
```diff
@@ -57,4 +57,47 @@ The gene expression data files available for download report all possible genes
...
+
+You will need to [install and load the `Seurat` package](https://satijalab.org/seurat/articles/install.html) to work with Seurat objects.
+
+For libraries that only contain RNA-sequencing data (i.e. do not have a CITE-seq library found in the `altExp` of the `SingleCellExperiment` object), you can use the following commands:
+
+```
+library(Seurat)
+
+# grab counts matrix from the SingleCellExperiment
+counts <- as.matrix(counts(sce))
```
jashapiro, line 75: I don't think this conversion is required? If Seurat can handle the sparse matrix directly, it would seem better to leave it that way for memory usage.
```suggestion
counts <- counts(sce)
```

docs/faq.md
```diff
@@ -57,4 +57,47 @@ The gene expression data files available for download report all possible genes
...
+
+The files available for download contain [`SingleCellExperiment` objects](http://bioconductor.org/books/3.13/OSCA.intro/the-singlecellexperiment-class.html). 
+If desired, these can be converted into Seurat objects. 
+
+You will need to [install and load the `Seurat` package](https://satijalab.org/seurat/articles/install.html) to work with Seurat objects.
+
+For libraries that only contain RNA-sequencing data (i.e. do not have a CITE-seq library found in the `altExp` of the `SingleCellExperiment` object), you can use the following commands:
+
+```
+library(Seurat)
```
jashapiro, line 72: I think we need both, or `counts()`  & `colData()` will not work. Or we could use `SingleCellExperiment::counts(sce)` etc. 
```suggestion
library(Seurat)
library(SingleCellExperiment)
```

docs/faq.md
```diff
@@ -57,4 +57,47 @@ The gene expression data files available for download report all possible genes
...
+
+# grab counts matrix from the SingleCellExperiment
+counts <- as.matrix(counts(sce))
+
+# create seurat object 
+seurat_object <- CreateSeuratObject(counts = counts, assay = "RNA",
+                                    project = "project_name",
+                                    # add colData from SingleCellExperiment
+                                    meta.data = as.data.frame(colData(sce)))
```
jashapiro, line 81: I would suggest we get this and rowData separately along with counts...

docs/faq.md
```diff
@@ -57,4 +57,47 @@ The gene expression data files available for download report all possible genes
...
+
+```
+library(Seurat)
+
+# grab counts matrix from the SingleCellExperiment
+counts <- as.matrix(counts(sce))
+
+# create seurat object 
+seurat_object <- CreateSeuratObject(counts = counts, assay = "RNA",
+                                    project = "project_name",
```
jashapiro, line 79: Does the project name have to be without spaces? Or should we use a (dummy) SCPCL id?

jashapiro (commented):
This looks good! It is a bit annoying that you have to pull parts out separately, and I think it may be worth adding this as a function in `scpcaTools` that we can point people to, but that is for another day.

My main question is how we want to handle pulling the various parts, because we should be consistent. For memory reasons, we probably don't want to encourage extra objects, but maybe we should have a section where we explain what each part is and what conversions are required or optional? (If we don't add `@meta.features`, are the gene IDs still there, for example.)

As I said earlier, I think you should definitely add the `readRDS()` step, as it explains where `sce` comes from. 

A few other minor comments and questions, mostly about whether we want to high

docs/faq.md
```diff
@@ -63,4 +63,68 @@ The gene expression data files available for download report all possible genes
...
+counts <- counts(sce)
+
+# colData and rowData will need to be converted to a data.frame to be used in the Seurat object
+cell_metadata <- as.data.frame(colData(sce))
+row_metadata <- as.data.frame(rowData(sce))
+
+# create seurat object 
+seurat_object <- CreateSeuratObject(counts = counts, 
+                                    assay = "RNA",
+                                    project = "SCPCL000000")
```
jashapiro, line 94: A few lines in here were duplicated later, where I think they make more sense (after explanation of why we need them). 

I also removed pulling a separate `counts` object, as that is the biggest part, so it would be better to encourage fewer copies of that matrix.

```suggestion
# create seurat object from the SCE counts matrix
seurat_object <- CreateSeuratObject(counts = counts(sce), 
                                    assay = "RNA",
                                    project = "SCPCL000000")
```

docs/faq.md
```diff
@@ -63,4 +63,68 @@ The gene expression data files available for download report all possible genes
...
+```
+The above code will only maintain information found in the original counts matrix from the `SingleCellExperiment`.
+Optionally, if you would like to keep the included cell and gene associated metadata during conversion to the Seurat object you can perform the below additional steps: 
+
+```
+# colData and rowData will need to be converted to a data.frame to be used in the Seurat object
```
jashapiro, line 100:
```suggestion
# convert colData and rowData to data.frame for use in the Seurat object
```

docs/faq.md
```diff
@@ -63,4 +63,68 @@ The gene expression data files available for download report all possible genes
...
+
+# add row metadata (rowData) from SingleCellExperiment to Seurat 
+seurat_object[["RNA"]]@meta.features <- row_metadata
+
+# add metadata from SingleCellExperiment to Seurat
+seurat_object@misc <- metadata(sce)
+```
+
+In order to convert `SingleCellExperiment` objects from libraries that have counts from both RNA-sequencing and CITE-seq, you can use the following additional commands.
+This adds a second assay containing the CITE-seq counts and associated feature data to an existing Seurat object:
```
jashapiro, line 115:
```suggestion
For `SingleCellExperiment` objects from libraries with both RNA-seq and CITE-seq data, you can use the following additional commands to add a second assay containing the CITE-seq counts and associated feature data:
```
```

File: docs/faq.md [hunk 9]
```diff
@@ -63,4 +63,68 @@ The gene expression data files available for download report all possible genes
...
+
+```
+# create assay object in Seurat from CITE-seq counts found in altExp(SingleCellExperiment)
+cite_counts <- counts(altExp(sce))
+cite_assay <- CreateAssayObject(counts = cite_counts)
```
jashapiro, line 120: I'm less sold on this "simpliciation" because it does create three levels of nesting, and this object is much smaller than the RNA-seq data, but parallelism makes me want to do it this way. (also, the same level of nesting appears in the next rows)

```suggestion
# create assay object in Seurat from CITE-seq counts found in altExp(SingleCellExperiment)
cite_assay <- CreateAssayObject(counts = counts(altExp(sce)))
```

docs/faq.md
```diff
@@ -63,4 +63,68 @@ The gene expression data files available for download report all possible genes
...
+
+# optional: convert rowData to data.frame for use in Seurat object
+cite_row_metadata <- as.data.frame(rowData(altExp(sce)))
+
+# optional: add row metadata (rowData) from altExp to assay 
+cite_assay@meta.features = cite_row_metadata
```
jashapiro, line 126: I think we can combine these two with one comment (also changing assignment operator)
```suggestion
# optional: add row metadata (rowData) from altExp to assay 
cite_row_metadata <- as.data.frame(rowData(altExp(sce)))
cite_assay@meta.features <- cite_row_metadata
```

PR #34, opened 2021-10-19
Add FAQ for using data in python instead of R 

### LINKED ISSUE #13: FAQ: How do I use RDS files with Python?
We were considering using the loom or h5ad format (https://github.com/AlexsLemonade/alsf-scpca/issues/122) to maximize compatibility with python. For our initial efforts, we will instead include an FAQ about how to use the RDS files with python where we include example code.

allyhawkins:
Closes #13. Here I added in a FAQ to address converting the RDS files to a usable format that can be read into python and used with `scanpy`. After a lot of trial and error and playing with HDF5 files and loom files, I think the most efficient solution is using `DropletUtils::write10xcounts()` to write the counts matrix into 10x format first and then reading into python (thanks Josh for starting to lead me down this route!). Here we can save each of the counts matrices separately for RNA-seq and CITE-seq and then use scanpy's `read_10x_mtx()` function and give it the path to the directory where the files that we just wrote out from R and it will read it in and create the `AnnData` object needed for use with `scanpy`. This takes < 30 seconds for writing the files and importing the files vs. the 15-30 minutes that it took to both write loom files and then read loom files back in, which was not efficient. 

The biggest downfall here vs suggesting the use of loom is that we lose the associated colData and rowData, but I don't think that's too much of a concern as I'm sure people going to the trouble of working with in it python might be recalculating much of that on their own. I also added in a link that provides some suggestions on how to save coldata and rowdata from R and then append it to an `AnnData` object. Other than that I think this is a fairly straight forward and efficient approach, but if we want to include the loom data option as well I can add that in. 

Also noting that this is stacked on #33.


docs/faq.md
```diff
@@ -116,4 +116,64 @@ cite_assay@meta.features <- cite_row_metadata
...
+
+```
+# extract the CITE counts matrix to be saved 
+cite_counts <- counts(altExp(sce))
+
+# write counts to 10X format
+DropletUtils::write10xCounts("SCPCL000000-cite", cite_counts, 
+                             barcodes = colnames(cite_counts),
+                             gene.id = rownames(cite_counts))
+
```
jashapiro, line 158: Same suggestion for simplification as above, but here I am adding the version 3 optional info that denotes this as antibody tags. If we want to do this, we should add `version = "3"` above as well.  (It looks like we _could_ combine the outputs into a single matrix, but I feel like keeping them separate still seems easier to understand & keep track of). 
```suggestion
# write CITE-seq counts to 10X format
DropletUtils::write10xCounts("SCPCL000000-cite", counts(altExp(sce)), 
                             barcodes = colnames(altExp(sce)),
                             gene.id = rownames(altExp(sce)),
                             gene.type = "Antibody Capture",
                             version = "3")

```

docs/faq.md
```diff
@@ -116,4 +116,64 @@ cite_assay@meta.features <- cite_row_metadata
...
+```
+
+These files can then be directly read into python using the [`scanpy` package](https://scanpy.readthedocs.io/en/stable/), creating an [`AnnData` object](https://anndata.readthedocs.io/en/latest/index.html).
```
jashapiro, line 161:
```suggestion
These files can then be directly read into Python using the [`scanpy` package](https://scanpy.readthedocs.io/en/stable/), creating an [`AnnData` object](https://anndata.readthedocs.io/en/latest/index.html).
```

docs/faq.md
```diff
@@ -116,4 +116,64 @@ cite_assay@meta.features <- cite_row_metadata
...
+anndata_object = sc.read_10x_mtx(rna_file_directory)
+
+cite_file_directory = "SCPCL000000-cite"
+cite_anndata = sc.read_10x_mtx(cite_file_directory)
+
+# append CITE-seq anndata to RNA-seq anndata
+anndata_object['CITE'] = cite_anndata.to_df()
+```
+
+It should be noted that in this conversion the `colData`, `rowData` and metadata that is found in the original `SingleCellExperiment` objects will not be obtained.
```
jashapiro, line 178:
```suggestion
It should be noted that in this conversion the `colData`, `rowData`, and metadata that are found in the original `SingleCellExperiment` objects will not be retained. 
```

docs/faq.md
```diff
@@ -116,4 +116,64 @@ cite_assay@meta.features <- cite_row_metadata
...
+
+cite_file_directory = "SCPCL000000-cite"
+cite_anndata = sc.read_10x_mtx(cite_file_directory)
+
+# append CITE-seq anndata to RNA-seq anndata
+anndata_object['CITE'] = cite_anndata.to_df()
+```
+
+It should be noted that in this conversion the `colData`, `rowData` and metadata that is found in the original `SingleCellExperiment` objects will not be obtained. 
+If you would like to maintain this data, we provide a [reference from the authors of scanpy](https://theislab.github.io/scanpy-in-R/#converting-from-r-to-python) discussing one approach to save pieces of the `SingleCellExperiment` object separately and add them into an `AnnData` object.
```
jashapiro, line 179: I'm splitting this up a bit, as I think there is more to the linked approach than just the conversion of metadata. I want to believe that for the simpler case, people going down this route will know how to write a data frame and read it in Python. 

I'm wondering if we should just link to the main part of that website (https://theislab.github.io/scanpy-in-R) , as the section you link to assumes you have read & set up earlier steps.  Or maybe we should link to https://theislab.github.io/scanpy-in-R/#the-r-markdown-approach ?

```suggestion
If you would like to include this data, you could write out each table separately and load them manually in Python.
Alternatively, you might be interested in this [reference from the authors of scanpy](https://theislab.github.io/scanpy-in-R/#converting-from-r-to-python) discussing a different approach to  conversion using Rmarkdown notebooks and the `reticulate` package to directly convert `SingleCellExperiment` object components to AnnData object components without writing files to disk.
```

docs/faq.md
```diff
@@ -116,4 +116,64 @@ cite_assay@meta.features <- cite_row_metadata
...
+# read in the RDS file to be converted
+sce <- readRDS("SCPCL000000_filtered.rds")
+
+# extract the counts matrix to be saved 
+rna_counts <- counts(sce)
+
+# write counts to 10X format and save to a folder named "SCPCL000000-rna"
+DropletUtils::write10xCounts("SCPCL000000-rna", rna_counts, 
+                             barcodes = colnames(rna_counts),
+                             gene.id = rownames(rna_counts))
```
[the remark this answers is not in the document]
jashapiro: > There is also a `chemistry` field

Update: this only matters if the output format is HDF5, which we are not using (and I would agree we should not here)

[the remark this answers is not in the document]
allyhawkins: I'm going to agree and vote no to the chemistry field if that is the case then.

allyhawkins:
I went ahead and added in a couple sentences on the version that we are using for writing out the files and that the files would not be compressed if using the default `version=2`. Let me know what you think.

docs/faq.md
```diff
@@ -116,4 +116,67 @@ cite_assay@meta.features <- cite_row_metadata
...
+```
+
+## What if I want to use Python instead of R? 
+
+We provide single-cell and single-nuclei gene expression data as RDS files, which must be opened in R to view the contents. 
+If you prefer to work in Python, there are a variety of ways of converting the count data to Python-compatible formats.
+We have found that one of the more efficient is conversion via the 10X format using [`DropletUtils::write10xCounts()`](https://rdrr.io/bioc/DropletUtils/man/write10xCounts.html). 
+Note that you will need to install the [`DropletUtils` package](https://www.bioconductor.org/packages/devel/bioc/html/DropletUtils.html) to use this function.
+
+This will output three files to a new directory:
```
jashapiro, line 128: Maybe we want to be a bit more specific here, as there are a number output formats we could apply. 

This maybe gets the info we need, without the details about .gz or not .gz?
```suggestion
When used as described below, `DropletUtils::write10xCounts()` will output three files to a new directory, following the format used by Cell Ranger 3.0 (and later):
```

PR #36, opened 2021-10-19
Add SCE contents

### LINKED ISSUE #14: QC-related metadata found in filtered RDS objects
We need to document what metadata we include in the filtered RDS objects that folks can use to make decisions about which cells and genes to include in their own analyses.

### LINKED ISSUE #27: Section: Gene expression file contents
As discussed in a meeting with @jashapiro, we should have a separate section to describe the contents of the `SingleCellExperiment` object that is in the RDS files for download. In this section we want to explain what is in the `colData` and `rowData`, specifically the meaning of each column name, and what is included in the metadata. 

In this section would be a good place to also include some instruction on commands that can be used to access different pieces of the SCE in R, such as using `colData(sce)` or `rowData(sce)`.

jashapiro:
Closes #27
Closes #14

Here I am filling out the section which describes the metadata. I divided it up into a section for the shared data for all samples, followed by separate ones describing the fields added for filtered data and CITE-seq data. I'm not 100% sure I love that organization: the other option I thought about was just having subsections within the three main components to describe the "sometimes" fields.

Other than that, look at each of the descriptions and code fragments; let me know if you think they are sufficient.

Note that I also changed the file name and main header, in anticipation of Bulk data coming down the line. I think I got all the links updated (RTD local build didn't complain), but let me know if you see anything I missed.


docs/sce_file_contents.md
```diff
@@ -0,0 +1,131 @@
...
+```r
+library(SingleCellExperiment)
+sce <- readRDS("SCPCL000000_filtered.rds")
+```
+
+## Common fields
+
+### Expression counts
+
+For both unfiltered and filtered single-cell and single-nucleotide experiments, the primary cell by gene RNA-seq expression count data is provided as the `"counts"` assay.
```
allyhawkins, line 20:
```suggestion
The `"counts"` assay of the `SingleCellExperiment` object for single-cell and single-nucleus experiments (both unfiltered and filtered) contains the primary RNA-seq expression data.
```
Something about this sentence wasn't sitting right with me (maybe partially because you accidentally said single-nucleotide 😆), but I also think you want to emphasize that the gene expression data is found in the `counts` assay, and less about the structure since the next few sentences explain that. So I reordered it and tried to get rid of some of the repetitiveness found in the next sentences.

docs/sce_file_contents.md
```diff
@@ -0,0 +1,131 @@
...
+Each column represents a cell or droplet, each row a gene.
+Column names are cell barcode sequences and row names are Ensembl gene IDs. 
+This data is stored as a sparse matrix that can be accessed with the following R code:
```
allyhawkins, line 23:
```suggestion
The data is stored as a sparse matrix, and each column represents a cell or droplet, each row a gene.
Column names are cell barcode sequences and row names are Ensembl gene IDs. 
The `counts` assay can be accessed with the following R code:
```
I think we should first say this is a matrix and then describe the contents of the matrix.

docs/sce_file_contents.md
```diff
@@ -0,0 +1,131 @@
...
+
+| Column name   | Contents                                                         |
+| ------------- | ---------------------------------------------------------------- |
+| `gene_symbol` | [HUGO](https://www.genenames.org) gene symbol, if defined        |
+| `mean`        | Mean count across all cells/droplets                             |
+| `detected`    | Number of cells in which the gene was detected (gene count > 0 ) |
+
+### Experiment metadata
+
+Metadata associated with data processing is included in the `metadata` slot as a list.
```
allyhawkins, line 68: Since most of these fields reference how the sample was processed should we add in a link back to the processing information here for more information on what each of these fields/ alevin-fry parameters actually mean?

docs/sce_file_contents.md
```diff
@@ -0,0 +1,131 @@
...
+| `alevinfry_version` | Version of `alevin-fry` used for mapping and quantification                                              |
+| `af_permit_type`    | `alevin-fry generate-permit-list` method used for filtering cell barcodes                                |
+| `af_resolution`     | `alevin-fry quant` resolution mode used                                                                  |
+| `usa_mode`          | Boolean indicating whether quantification was done using `alevin-fry` USA mode                           |
+| `af_num_cells`      | Number of cells reported by `alevin-fry`                                                                 |
+| `transcript_type`   | Transcripts included in gene counts: `spliced` for single-cell samples and `unspliced` for single-nuclei |
+
+## Additional fields for filtered cells
+
+After {ref}`filtering for empty droplets <processing_information:Filtering cells>`, some additional fields are added to the `SingleCellExperiment` object.
```
allyhawkins, line 90:
```suggestion
After {ref}`filtering and removing empty droplets <processing_information:Filtering cells>`, some additional fields are added to the `SingleCellExperiment` object.
```
I don't know if this is the best way to fix this but `filtering for empty droplets` made it sound like we were selecting for those to me, so I'm trying this to see if it helps clarify it.

docs/sce_file_contents.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Additional fields for CITE-seq samples
+
+CITE-seq data, when present, is included within the `SingleCellExperiment` as an "Alternative Experiment" named `"CITEseq"` , which can be accessed with the following command:
+
+```r
+altExp(sce, "CITEseq")
+```
+
+Within this, the main data matrix is again found in the `counts` assay, with each column corresponding to a cell or droplet (in the same order as the parent `SingleCellExperiment`) and each row corresponding to an antibody derived tag (ADT).
```
allyhawkins, line 103:
```suggestion
Within this, the main expression matrix is again found in the `counts` assay, with each column corresponding to a cell or droplet (in the same order as the parent `SingleCellExperiment`) and each row corresponding to an antibody derived tag (ADT).

To access the counts matrix, specifically for CITE-seq data, you can use the following command: 

```r
cite_counts <- counts(altExp(sce, "CITEseq"))
```
I think it might be helpful to show here how you would do the same things in parallel to what we showed with the RNA-seq data. I know its repetitive but it could be helpful to people who don't have as much experience with working with multi-modal data in particular.

allyhawkins (commented):
So after reading this through a few times and also sleeping on it, I'm not 100% sold on the organization. I like keeping the CITE-seq stuff separately because I do think that almost is an entirely different piece and makes sense to all be together. But I don't know how I feel about the additional fields for filtered objects being separate as I think most people will be using the filtered RDS files so its helpful to have it in the main body. I think it could be helpful to include a sentence after the `colData` section for the unfiltered that says we added the `prob_compromised` in for filtered sces. And then same thing for the metadata, where theres just an additional note at the bottom rather than a whole other section. 

Other than that I just made a few wording suggestions you can keep or not and also think we might want to add  a link in the metadata section to the processing information section of the docs. I think this is coming together nicely though and I'm glad we decided to do another section entirely to walk through this! I hope people find it helpful.

docs/sce_file_contents.md
jashapiro: Agree with the idea here, but I don't think of this as expression? I guess protein expression? I'll probably use expression unless I can come up with something better.

docs/sce_file_contents.md
allyhawkins: Oh you are right I just looked through some of the other docs and we are using single-nuclei. I think we should stick with that as I'm pretty sure that's also how 10X references it.

docs/sce_file_contents.md
```diff
@@ -0,0 +1,127 @@
...
+| `alevinfry_version` | Version of `alevin-fry` used for mapping and quantification                                                                    |
+| `af_permit_type`    | `alevin-fry generate-permit-list` method used for filtering cell barcodes                                                      |
+| `af_resolution`     | `alevin-fry quant` resolution mode used                                                                                        |
+| `usa_mode`          | Boolean indicating whether quantification was done using `alevin-fry` USA mode                                                 |
+| `af_num_cells`      | Number of cells reported by `alevin-fry`                                                                                       |
+| `transcript_type`   | Transcripts included in gene counts: `spliced` for single-cell samples and `unspliced` for single-nuclei                       |
+| `miQC_model`        | The model object that `miQC` fit to the data and was used to calculate `prob_compromised`. Only present for `filtered` objects |
+
+
+## Additional SingleCellExperiment components for CITE-seq samples
```
allyhawkins, line 91: Do we want to put write SingleCellExperiment in code format as you are doing with the first header? I don't particularly have a preference, but we should be consistent on which way we choose to do it.

allyhawkins (approved):
I think this looks good and I like the incorporation of what is in the filtered sce's to the `colData` and metadata. I think the headers are okay, I'm having trouble coming up with an alternative, but I just left one comment about consistency of the formatting around `SingleCellExperiment`. Otherwise looks good.

docs/sce_file_contents.md
jashapiro: I was going back and forth and forgot I did use code format before. I'll use that everywhere.

PR #41, opened 2021-10-27
Add processing information for bulk RNA-sequencing

### LINKED ISSUE #9: Processing Information for Bulk RNA-seq
We will need to add the details about processing of bulk RNA-sequencing to the processing information section.

allyhawkins:
Closes #9. This PR adds a section to the processing information section of the docs about how we performed processing of bulk RNA-sequencing files. I chose to break it up into two sections, one for the preprocessing with `fastp` and then one for mapping and quantification with `salmon`. As with the previous sections, I only included discussion about any extra parameters that vary from the default usage. 

I chose not to include a separate section for discussion of the index as we did with the splici index, since it didn't seem to need any extra justification as it's already published to use a decoy sequence, but let me know if I should expand on this point more. 

Because we are not including bulk RNA-sequencing in the beta version, we shouldn't merge this yet to avoid confusion, but I wanted to get this written while this was fresher in my mind. Because of this, I'll go ahead and leave it in draft form for now until we're ready to include this.


allyhawkins:
I went ahead and added a `(Data Available Soon)` to the heading in the bulk processing section as to minimize confusion on having this in but not having this data available. I wasn't 100% on this wording so definitely open to changing how we say that the data is coming soon. We will just have to remember to remove it. 

I also added in a short section in the download files section on the bulk file thats included if a project has bulk RNA-sequencing. I put it right below the illustration on downloading by project but before we discuss downloading by samples as this seemed to me to make the most sense rather than having a separate section on it later on. But I'm open to that option if others think that makes more sense.

docs/download_files.md
```diff
@@ -7,6 +7,10 @@ The folder structure within the zip file is determined by whether individual sam
 ## Download folder structure for project downloads:
 ![docs-downloads-project](https://user-images.githubusercontent.com/15315514/143308420-a3cca10d-814f-4c52-b934-98d5e9cef1c5.png)
 
+If a project contains bulk RNA-sequencing samples, a single tab-separated values file, `SCPCP000000_bulk_quant.tsv` will be included in the download. 
+The contents of this file include a gene by sample matrix containing raw gene expression counts data determined using Salmon.
```
jashapiro, line 11: This file does not have anything but the gene by sample matrix, correct? In my mind "contents include" implies that there may be other things as well.

I also clarified the row/column contents, but I don't know if that is necessary.
```suggestion
This file contains a gene by sample matrix (each row a gene, each column a sample) containing raw gene expression counts quantified by Salmon.
```

jashapiro (commented):
This looks good. The overall content is all there as far as I can tell. I had a number of smallish style suggestions, with just a bit of hopefully justified simplification.

allyhawkins:
Thanks for taking a look at this @jashapiro. I liked your suggestions so went ahead and incorporated those and then also made sure that we are uniformly using `RNA-seq` rather than mixing it with `RNA-sequencing` throughout the docs. I found one other place where we had used `RNA-sequencing` so changed that to match as well.

jashapiro (approved):
LGTM!

PR #42, opened 2021-11-24
Added download illustrations

dvenprasad:
Related to #6 

Added illustrations to the section.

### DIFF
--- a/docs/download_files.md
+++ b/docs/download_files.md
@@ -4,21 +4,21 @@ The ScPCA Portal download packages include gene expression data, a QC report, an
 These files are delivered as a zip file.
 The folder structure within the zip file is determined by whether individual samples or all samples associated with a project are selected for download.  
 
-## Download folder structure for project downloads: 
-_add in illustration of file structure for project_ 
+## Download folder structure for project downloads:
+![docs-downloads-project](https://user-images.githubusercontent.com/15315514/143308420-a3cca10d-814f-4c52-b934-98d5e9cef1c5.png)
 
-## Download folder structure for individual sample downloads: 
-_add in illustration of file structure for sample downloads_ 
+## Download folder structure for individual sample downloads:
+![docs-downloads-sample](https://user-images.githubusercontent.com/15315514/143308420-a3cca10d-814f-4c52-b934-98d5e9cef1c5.png)
 
-Sample folders (indicated by the `SCPCS` prefix) contain the files for all libraries (`SCPCL` prefix) derived from that biological sample. 
-Most samples only have one library that has been sequenced. 
+Sample folders (indicated by the `SCPCS` prefix) contain the files for all libraries (`SCPCL` prefix) derived from that biological sample.
+Most samples only have one library that has been sequenced.
 See the {ref}`FAQ section about samples and libraries <faq:What is the difference between samples and libraries?>` for more information.
 
 The files associated with each library are (example shown for a library with ID `SCPCL000000`):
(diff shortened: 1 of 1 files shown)


docs/download_files.md
```diff
@@ -4,21 +4,21 @@ The ScPCA Portal download packages include gene expression data, a QC report, an
 These files are delivered as a zip file.
 The folder structure within the zip file is determined by whether individual samples or all samples associated with a project are selected for download.  
 
-## Download folder structure for project downloads: 
-_add in illustration of file structure for project_ 
+## Download folder structure for project downloads:
+![docs-downloads-project](https://user-images.githubusercontent.com/15315514/143308420-a3cca10d-814f-4c52-b934-98d5e9cef1c5.png)
```
allyhawkins, line 8: I think this is the same image as below? And when you go to these links neither of them are the correct image that you had originally posted on the issue?

dvenprasad: Ah corrected the links

No, I updated the it to reflect the structure of the download folder I got from the portal.

allyhawkins: Okay so this image (and presumably the portal then) does not include the `SCPCL000_metadata.json` for every library as we stated in the below docs? 
I believe the intent that @jashapiro had in mind was to include that file in the download but we should confirm that is correct and if so then see if @kurtwheeler can include it. Otherwise we need to update the rest of the docs to remove that file from the expected contents.

kurtwheeler: Oh I didn't realize we wanted to include those files. I thought we were pretty much rewriting all that metadata to the `libraries_metadata.csv` file anyway?

jashapiro: I was kind of agnostic on the question, generally. I think my main thought was that for the individual sample downloads it didn't necessarily make sense to rewrite the file, and that json is potentially _more_ human-readable for a single sample. But this is not a particularly strongly-held view.

allyhawkins: I do believe the only things we are currently grabbing from the json file and outputting to the metadata that is included in the `libraries_metadata.csv` is the `seq_unit`, `technology`, and number of cells right? There are other pieces of information that are included inside the R objects that are taken from the json file, but then the only way to view those is to open up the R object. I think to Josh's point that having it in json format provides a slightly more upfront and human-readable version of metadata that is accessible outside of having to load the files into R. So yes everything is presented somewhere else in the download with the way that it is currently, it's just a question of how we want to present it and if we want it to be available in multiple forms. If no one feels strongly about having it then we should get rid of it because it is duplicating information, we just will need to make other changes to the docs.

allyhawkins (approved):
Going to go ahead and approve this and then will file an issue for the changes that needed to be made based on the download contents.

docs/download_files.md
jashapiro: I missed this comment earlier, and it did remind me of my earlier reasoning, which was that getting some of the details of software versions etc. might be something that people want to look at without having to load a large RDS file. That said, that info _is_ in the html report, so the use case here would be somebody trying to get that info programmatically, not just peeking at a couple samples. 

So I think leaving it out is fine for now, though I wonder if @dvenprasad  may want to ask if people during testing (or after launch) if they want that info (or other info that is in the html report) in a more lightweight machine-readable form.

dvenprasad: There's a section in the eval where I ask them to look at download files and tell me what they think, and talk about what their next steps are. This question can fit well there.

PR #47, opened 2022-01-14
Add documentation on processing spatial transcriptomics libraries

### LINKED ISSUE #8: Processing Information for Spatial Transcriptomics
The processing information section will need to include the information about how we process spatial transcriptomics data (once that module is added to the workflow).

allyhawkins:
Closes #8. I added in a section in the `processing_information.md` file for spatial transcriptomic libraries that discusses how they were processed using Space Ranger. I kept things relatively simple here and directed users to the documentation on Space Ranger since we aren't doing anything other than using the default parameters for Space Ranger and there is documentation on the 10X site on how things are done. Let me know if I should add in more detail here and take a different approach. 

Also I added in one sentence that Alevin-fry did not have the capability to perform tissue detection but only can perform the gene expression quantification, so therefore we are using Space Ranger that can do both together. I'm definitely open with other ways of framing why we chose Space Ranger over Alevin-fry but was worried it would get complicated with explaining that the quantification in selective alignment wasn't recovering the same counts per spot but that can be solved by using Sketch. I thought better to avoid that. 

The next thing to do here would be to add in a section on the actual output files that you obtain in the spatial libraries, but I think we want to make a new issue and wait on adding that until we know how things will be going in and out of the portal. We will probably then just want to either add a section to the `download_files.md` that has the structure of the files that are included or create another section like we did with `sce_file_contents.md`. 

This PR also is going to have conflicts with #41, which adds the bulk processing information to the end of `processing_information.md`. I could be swayed either way, but I think we want to actually put all single-cell/spot related stuff together and then have the bulk at the end so this should go before the bulk (unless others have strong other opinions about order) and we can address those conflicts after #41 is merged.


docs/processing_information.md
```diff
@@ -65,3 +65,17 @@ When normalizing these two count matrices to the same set of cells, we chose to
...
+
+**_Data Available Soon_**
+
+### Mapping and quantification using Space Ranger
+
+Processing spatial transcriptomics libraries requires two steps - gene expression quantification and tissue detection. 
+Currently, Alevin-fry is not capable of tissue detection, but only reports gene expression. 
+Therefore, we utilized [10X Genomics' Space Ranger](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/what-is-space-ranger) to obtain both gene expression and spatial information. 
+[`spaceranger count`](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/using/count) takes a microscopic slide image and FASTQ files as input and performs alignment, tissue detection, and spot barcode and UMI counting. 
+In contrast to Alevin-fry which maps reads to a [reference transcriptome index](#reference-transcriptome-index), Space Ranger aligns transcript reads to the reference genome using STAR ([Dobin _et al_., 2012](https://doi.org/10.1093/bioinformatics/bts635)).
```
jashapiro, line 79:
```suggestion
In contrast to Alevin-fry, which maps reads to a [reference transcriptome index](#reference-transcriptome-index), Space Ranger aligns transcript reads to the reference genome using STAR ([Dobin _et al_., 2012](https://doi.org/10.1093/bioinformatics/bts635)).
```

docs/processing_information.md
```diff
@@ -65,3 +65,17 @@ When normalizing these two count matrices to the same set of cells, we chose to
...
+
+## Spatial Transcriptomics
+
+**_Data Available Soon_**
+
+### Mapping and quantification using Space Ranger
+
+Processing spatial transcriptomics libraries requires two steps - gene expression quantification and tissue detection. 
+Currently, Alevin-fry is not capable of tissue detection, but only reports gene expression. 
+Therefore, we utilized [10X Genomics' Space Ranger](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/what-is-space-ranger) to obtain both gene expression and spatial information.
```
jashapiro, line 77: Merging these two sentences, in the hopes of a bit less redundancy. I also want to sort of imply that AF shouldn't really be expected to do tissue id. (What I would really want is a `spaceranger imageprocess` command, but I'm not gonna get that.)
```suggestion
In the absence of independent tissue detection methods to use with Alevin-fry, we used [10X Genomics' Space Ranger](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/what-is-space-ranger) to obtain both gene expression and spatial information. 
```

docs/processing_information.md
```diff
@@ -65,3 +65,17 @@ When normalizing these two count matrices to the same set of cells, we chose to
...
+## Spatial Transcriptomics
+
+**_Data Available Soon_**
+
+### Mapping and quantification using Space Ranger
+
+Processing spatial transcriptomics libraries requires two steps - gene expression quantification and tissue detection. 
+Currently, Alevin-fry is not capable of tissue detection, but only reports gene expression. 
+Therefore, we utilized [10X Genomics' Space Ranger](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/what-is-space-ranger) to obtain both gene expression and spatial information. 
+[`spaceranger count`](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/using/count) takes a microscopic slide image and FASTQ files as input and performs alignment, tissue detection, and spot barcode and UMI counting.
```
jashapiro, line 78: I simplified the end of this sentence, and flipped the order of the input files to match the "expression, image" order that we have throughout this section. 
```suggestion
[`spaceranger count`](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/using/count) takes FASTQ files and a microscopic slide image as input and performs alignment, quantification, and tissue detection for each spot. 
```

jashapiro (approved):
This looks good! I have a few minor wording edits, but not much more.

I definitely support leaving out the discussion of alevin-fry details... We would then have to explain why it is okay for single cell but not for spatial, which really gets us into a whole mess! (this might be an FAQ question, but I might wait for someone to actually ask).

For merging, I agree with you that spatial should come before bulk. (I also don't think it matters too much.)

PR #54, opened 2022-02-15
Add information about hard UMI cutoff for filtering

### LINKED ISSUE #53: Add note about using hard cutoff for filtering in docs
Per incorporation of https://github.com/AlexsLemonade/scpcaTools/pull/91, we now have the option for libraries that fail `emptyDropsCellRanger` or `emptyDrops` to be filtered using a hard cutoff. There is a warning that is printed in the QC report if this happens, but we probably want to add a line to the docs that this is an option and that we've chosen a default cutoff of UMI >= 100.

allyhawkins:
Closes #53. Here, I am adding a short note in the filtering section of the processing information to note that if a library does not meet the assumption of `emptyDropsCellRanger`, that it will be filtered using a hard cutoff instead. 

I also added in the additional fields in the metadata `filtering_method` and `umi_cutoff` to the `sce_file_contents` page. I chose to keep everything in the metadata for the sce as one large table, but thought about the idea of breaking out the three pieces of metadata that are only present for the filtered data and include in a small separate table below the first table (similar to how we do the `cell metrics` section). If reviewers like that idea, I can break out the table, but I wasn't sure if this was completely necessary.

### DIFF
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -53,6 +53,9 @@ This function more closely mimics the filtering performed in Cell Ranger than it
 We consider droplets with an FDR less than or equal to 0.01 to be cell-containing droplets. 
 Only cells that pass this FDR threshold are included in the filtered counts matrix.
 
+In the event that the cells do not meet the assumptions met in `DropletUtils::emptyDropsCellRanger` and are unable to be filtered using this method (e.g. have a lower than expected number of droplets), droplets were filtered using a hard cutoff in the total number of UMI. 
+For these libraries, only droplets containing greater than 100 UMI are included in the filtered counts matrix.
+
 ## CITE-seq quantification
 
 CITE-seq libraries with reads from antibody-derived tags (ADTs) were also quantified using  [`salmon alevin`](https://salmon.readthedocs.io/en/latest/alevin.html) and [`alevin-fry`](https://alevin-fry.readthedocs.io/en/latest/), rounded to integer values.
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -53,7 +53,7 @@ This function more closely mimics the filtering performed in Cell Ranger than it
 We consider droplets with an FDR less than or equal to 0.01 to be cell-containing droplets. 
 Only cells that pass this FDR threshold are included in the filtered counts matrix.
 
-In the event that the cells do not meet the assumptions met in `DropletUtils::emptyDropsCellRanger` and are unable to be filtered using this method (e.g. have a lower than expected number of droplets), droplets were filtered using a hard cutoff in the total number of UMI. 
+In the event that the cells do not meet the assumptions met in `DropletUtils::emptyDropsCellRanger` and are unable to be filtered using this method (e.g. have a lower than expected number of droplets), droplets were filtered based on total UMI count. 
 For these libraries, only droplets containing greater than 100 UMI are included in the filtered counts matrix.
 
 ## CITE-seq quantification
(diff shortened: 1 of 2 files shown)


docs/processing_information.md
```diff
@@ -53,6 +53,9 @@ This function more closely mimics the filtering performed in Cell Ranger than it
 We consider droplets with an FDR less than or equal to 0.01 to be cell-containing droplets. 
 Only cells that pass this FDR threshold are included in the filtered counts matrix.
 
+In the event that the cells do not meet the assumptions met in `DropletUtils::emptyDropsCellRanger` and are unable to be filtered using this method (e.g. have a lower than expected number of droplets), droplets were filtered based on total UMI count.
```
jashapiro, line 56: I wanted to simplify the structure of this a bit. Since the next sentence contains the filtering we do in the case of failure, I eliminated that clause.
```suggestion
For some libraries, `DropletUtils::emptyDropsCellRanger()` may fail due to low numbers of droplets with reads or other violations of its assumptions.
```

docs/processing_information.md
```diff
@@ -53,6 +53,9 @@ This function more closely mimics the filtering performed in Cell Ranger than it
 We consider droplets with an FDR less than or equal to 0.01 to be cell-containing droplets. 
 Only cells that pass this FDR threshold are included in the filtered counts matrix.
 
+In the event that the cells do not meet the assumptions met in `DropletUtils::emptyDropsCellRanger` and are unable to be filtered using this method (e.g. have a lower than expected number of droplets), droplets were filtered based on total UMI count. 
+For these libraries, only droplets containing greater than 100 UMI are included in the filtered counts matrix.
```
jashapiro, line 57: We are including 100 UMI droplets, right?
```suggestion
For these libraries, only droplets containing at least 100 UMI are included in the filtered counts matrix.
```

docs/sce_file_contents.md
```diff
@@ -93,7 +93,8 @@ expt_metadata <- metadata(sce)
 | `tech_version`      | A string indicating the technology and version used for the single-cell library, such as 10Xv2, 10Xv3, or 10Xv3.1              |
 | `transcript_type`   | Transcripts included in gene counts: `spliced` for single-cell samples and `unspliced` for single-nuclei                       |
 | `miQC_model`        | The model object that `miQC` fit to the data and was used to calculate `prob_compromised`. Only present for `filtered` objects |
-
+| `filtering_method`  | Which method was used for filtering, one of `emptyDrops`, `emptyDropsCellRanger`, or `UMI cutoff`. Only present for `filtered` objects |
```
jashapiro, line 96:
```suggestion
| `filtering_method`  | The method used for cell filtering. One of `emptyDrops`, `emptyDropsCellRanger`, or `UMI cutoff`. Only present for `filtered` objects |
```

docs/sce_file_contents.md
```diff
@@ -93,7 +93,8 @@ expt_metadata <- metadata(sce)
...
+| `umi_cutoff`        | Only present if the `filtering_method` is `UMI cutoff`. The minimum number of UMI per cell used as a threshold for filtering. Only present for `filtered` objects.  |
```
jashapiro, line 97: Eliminating the double "only". Also removed the terminal period because we seem to be doing that through these tables. 
```suggestion
| `umi_cutoff`        | The minimum UMI count per cell used as a threshold for filtering. Only present for `filtered` objects where the `filtering_method` is `UMI cutoff` |
```

jashapiro (commented):
Thanks for doing this! I had a number of smaller comments about wording, see below.

allyhawkins:
Thanks for taking a look at this and for helping with cleaning up the wording! I incorporated your changes so it's ready for another look.

jashapiro (approved):
LGTM!

PR #56, opened 2022-02-24
Add bulk metadata file to download section 

### LINKED ISSUE #55: Add `bulk_metadata.tsv` file to download section for bulk data
We do not mention the `bulk_metadata.tsv` file that is included in the downloads for projects with bulk RNA-seq data. We should update the `download_files` section to reflect that if a project has bulk RNA-seq users will see both the metadata and the quants files.

allyhawkins:
Closes #55. Here I am adding in a short section on the `bulk_metadata.tsv` file that is included in downloads for projects with bulk RNA-sequencing. I chose to just add one sentence about the metadata file instead of going into detail about every single-column because I wasn't quite sure we needed that level of detail. If others feel differently then I'm happy to add in a table to the section on `metadata` at the bottom of the page with the details for each column. 

Based on what we add here, I also wanted to make the associated changes for https://github.com/AlexsLemonade/scpca-portal/issues/144, copying over the section we have here about if a project contains bulk files. Based on any wording changes that may be made here, I will then make the necessary changes to address https://github.com/AlexsLemonade/scpca-portal/issues/144 so we don't have to open PR's with similar wording.


docs/download_files.md
```diff
@@ -7,8 +7,10 @@ The folder structure within the zip file is determined by whether individual sam
 ## Download folder structure for project downloads:
 ![docs-downloads-project](https://user-images.githubusercontent.com/15315514/143308420-a3cca10d-814f-4c52-b934-98d5e9cef1c5.png)
 
-If a project contains bulk RNA-seq samples, a single tab-separated values file, `SCPCP000000_bulk_quant.tsv` will be included in the download. 
-This file contains a gene by sample matrix (each row a gene, each column a sample) containing raw gene expression counts quantified by Salmon.
+If a project contains bulk RNA-seq samples, two tab-separated value files, `SCPCP000000_bulk_quant.tsv` and `SCPCP000000_bulk_metadata.tsv`, will be included in the download. 
+The `SCPCP000000_bulk_quant.tsv` file contains a gene by sample matrix (each row a gene, each column a sample) containing raw gene expression counts quantified by Salmon.
+The `SCPCP000000_bulk_metadata.tsv` file contains associated metadata, such as the type of RNA-seq, total number of reads, and processing information for each sample.
```
jashapiro, line 12: Are the fields largely the same as for the single cell data? Can we combine the descriptions for those two tables into a single section? 

More generally, are we copying the sample metadata into this table at import as well, or is this staying separate? That is a question for @davidsmejia I suppose. Maybe also for @dvenprasad, as to what we _should_ do. 

For more context, the bulk sample metadata we provide for import has library information related to the assay, but does not contain sample information such as disease type. We _do_ merge in that information in the `library_metadata.tsv` file, as described below in the **Metadata** section, but that file only includes the single cell libraries.

I think it makes sense to keep the two data tables separate, despite the fact that many fields will be similar if we do merge in metadata, but I wonder if we might want to rename `library_metadata.tsv` below to match the naming of these files, i.e., `SCPCP000000_sc_metadata.tsv`.

allyhawkins: > Are the fields largely the same as for the single cell data? Can we combine the descriptions for those two tables into a single section?

In looking at a test download of the single-cell data and the fields provided in `libraries_metadata.csv` in compared to what we provide in the `bulk_metadata.tsv`, the `bulk_metadata.tsv` columns are mostly all found in the single-cell metadata. The single-cell metadata has some additional single-cell specific columns like filtered cell counts and then the additional metadata for that sample like disease, age, etc. 
I do want to note that there are some additional columns that I noticed that are not currently accounted for in the docs, `salmon_version`, `alevin_fry_version`, `unfiltered_cells`, `total_reads`, `mapped_reads`, `transcript_type`, and `cell_count` (? this also appears to duplicate filtered cell count). It might be worth adding some of these to our table in the docs. 

I think we could add a sentence in the metadata section stating that the bulk file only contains processing information and does not contain columns related to sample metadata. Unless we think it would be important to have it repeated in both places? 

> I think it makes sense to keep the two data tables separate, despite the fact that many fields will be similar if we do merge in metadata, but I wonder if we might want to rename library_metadata.tsv below to match the naming of these files, i.e., SCPCP000000_sc_metadata.tsv.

I do like the idea of renaming the single-cell metadata to match the bulk metadata.

davidsmejia: > More generally, are we copying the sample metadata into this table at import as well, or is this staying separate? That is a question for @davidsmejia I suppose. Maybe also for @dvenprasad, as to what we should do.

Will be separate, nested inside of the project folder.

I renamed the files to: 
`libraries_metadata.csv` => `single_cell_metadata.csv` 
`SCPCP000000_bulk_metadata.tsv` =>  `bulk_metadata.tsv`
`SCPCP000000_bulk_quant.tsv` => `bulk_quant.tsv`

They will be nested in the project folder so I think we should either consistently prefix with project id or omit it.

allyhawkins: > I renamed the files to:
libraries_metadata.csv => single_cell_metadata.csv
SCPCP000000_bulk_metadata.tsv => bulk_metadata.tsv
SCPCP000000_bulk_quant.tsv => bulk_quant.tsv

I like this change. If @jashapiro is onboard then I'll update the docs to reflect these changes but we will also need to update the screenshots.

jashapiro: I like it, but for one thing... `single_cell_metadata.csv` will come after `SCPCL` alphabetically, so those files won't sort to the top(or together). But I don't think that is worth modifying the name for.

allyhawkins:
@jashapiro I believe this is ready for another look. I went ahead and updated the file names and will then file an issue to update the screenshots with the new file names. 

I also tried something with the metadata explanation where I'm linking out to the descriptions of the [metadata that is actually in the SCE object](https://scpca.readthedocs.io/en/latest/sce_file_contents.html#experiment-metadata), since much of that metadata is getting pulled into the `single_cell_metadata.tsv`. I started to add it to the existing table, but much of it is repeated there so it seemed redundant. 

I then added a sentence about the bulk metadata containing all of the processing related metadata and none of the disease specific metadata, noting the columns that are removed. Let me know if this is not clear to you and if you would prefer a different approach.

docs/download_files.md
allyhawkins: I don't think it matters theoretically, but it's probably a good idea to convert them all to the same format for consistency. Changing the bulk metadata to `csv` would require updates to the workflow and re-running the projects since that is generated through `scpca-nf`, which shouldn't be a huge hassle but is not minimal effort. I'm assuming that changing the `libraries_metadata.csv` to `tsv` would also require changes to creating that file on your end though @davidsmejia so the question is which one is the least disruptive and how important is it to change it.

davidsmejia: I can change `libraries_metadata.csv` -> `single_cell_metadata.csv` -> `single_cell_metadata.tsv`

allyhawkins: My brain clearly was not working, I forgot both changes would be pretty easy. But that works 👍

docs/download_files.md
```diff
@@ -23,7 +25,7 @@ The files associated with each library are (example shown for a library with ID
 - A filtered counts file: `SCPCL000000_filtered.rds`,
 - A quality control report: `SCPCL000000_qc.html`,
 
-Every download also includes a single `libraries_metadata.csv` file containing metadata for all libraries included in the download.
+Every download also includes a single `single_cell_metadata.tsv` file containing metadata for all libraries included in the download.
```
jashapiro, line 28: I feel like this line should be in both the individual sample and project sections. It seems conspicuously absent from the project information. Or perhaps it should be at ~ line 6, since it applies to both.

docs/download_files.md
```diff
@@ -7,8 +7,10 @@ The folder structure within the zip file is determined by whether individual sam
 ## Download folder structure for project downloads:
 ![docs-downloads-project](https://user-images.githubusercontent.com/15315514/143308420-a3cca10d-814f-4c52-b934-98d5e9cef1c5.png)
 
-If a project contains bulk RNA-seq samples, a single tab-separated values file, `SCPCP000000_bulk_quant.tsv` will be included in the download. 
-This file contains a gene by sample matrix (each row a gene, each column a sample) containing raw gene expression counts quantified by Salmon.
+If a project contains bulk RNA-seq samples, two tab-separated value files, `bulk_quant.tsv` and `bulk_metadata.tsv`, will be included in the download.
```
jashapiro, line 10: minor wording change
```suggestion
If a project contains bulk RNA-seq data, two tab-separated value files, `bulk_quant.tsv` and `bulk_metadata.tsv`, will be included in the download. 
```

docs/download_files.md
```diff
@@ -7,8 +7,10 @@ The folder structure within the zip file is determined by whether individual sam
 ## Download folder structure for project downloads:
 ![docs-downloads-project](https://user-images.githubusercontent.com/15315514/143308420-a3cca10d-814f-4c52-b934-98d5e9cef1c5.png)
 
-If a project contains bulk RNA-seq samples, a single tab-separated values file, `SCPCP000000_bulk_quant.tsv` will be included in the download. 
-This file contains a gene by sample matrix (each row a gene, each column a sample) containing raw gene expression counts quantified by Salmon.
+If a project contains bulk RNA-seq samples, two tab-separated value files, `bulk_quant.tsv` and `bulk_metadata.tsv`, will be included in the download. 
+The `bulk_quant.tsv` file contains a gene by sample matrix (each row a gene, each column a sample) containing raw gene expression counts quantified by Salmon.
+The `bulk_metadata.tsv` file contains associated metadata for all samples with bulk RNA-seq.
```
jashapiro, line 12:
```suggestion
The `bulk_metadata.tsv` file contains associated metadata for all samples with bulk RNA-seq data.  
```

docs/download_files.md
```diff
@@ -68,3 +70,8 @@ The `libraries_metadata.csv` file is a comma-separated table with one row per li
 
 Additional metadata may also be included, specific to the disease type and experimental design of the project.
 Examples of this include treatment or outcome.
+Metadata pertaining to processing will also be available in this table and inside of the `SingleCellExperiment` object.
+See the {ref}`Experiment metadata <sce_file_contents:experiment metadata>` section for more information on metadata columns that can be found in this file as well as inside the `SingleCellExperiment` object.
+
+For projects with bulk RNA-seq, the `bulk_metadata.tsv` file will be included for project downloads.
```
jashapiro, line 76: Trying to be consistent with my syggestions.
```suggestion
For projects with bulk RNA-seq data, the `bulk_metadata.tsv` file will be included for project downloads. 
```

docs/download_files.md
```diff
@@ -68,3 +70,8 @@ The `libraries_metadata.csv` file is a comma-separated table with one row per li
 
 Additional metadata may also be included, specific to the disease type and experimental design of the project.
 Examples of this include treatment or outcome.
+Metadata pertaining to processing will also be available in this table and inside of the `SingleCellExperiment` object.
+See the {ref}`Experiment metadata <sce_file_contents:experiment metadata>` section for more information on metadata columns that can be found in this file as well as inside the `SingleCellExperiment` object.
+
+For projects with bulk RNA-seq, the `bulk_metadata.tsv` file will be included for project downloads. 
+This file will contain all of the fields found in the `single_cell_metadata.tsv` related to processing the sample, but will not contain patient or disease specific metadata (e.g. `age`, `sex`, `diagnosis`, `subdiagnosis`, `tissue_location`, or `disease_timing`).
```
jashapiro, line 77: Removing "all of" in case it is not always true.
```suggestion
This file will contain fields equivalent to those found in the `single_cell_metadata.tsv` related to processing the sample, but will not contain patient or disease specific metadata (e.g. `age`, `sex`, `diagnosis`, `subdiagnosis`, `tissue_location`, or `disease_timing`).
```

docs/download_files.md
jashapiro: I actually wonder if all of lines 19-26 (starting with "Sample folders (indicated by the `SCPCS` prefix)" up to " A quality control report") should be in that initial section, since those contents apply to both kinds of downloads?

jashapiro (commented):
This looks good, I just had some remaining organization thoughts.  Also, I am wondering if we want to wait for updated images to merge this in?

allyhawkins:
I incorporated your suggestions @jashapiro and went ahead and reorganized the section slightly so that the beginning explains all files that are included in either type of download and then we show the screenshots that display the differences in downloading by project and by sample. Was this what you were thinking in https://github.com/AlexsLemonade/scpca-docs/pull/56#discussion_r817806138. 

I also filed #57 to update the screenshots so we can wait to merge this until the screenshots are integrated.

jashapiro (approved):
LGTM!

Issue #69, opened 2022-04-11
Add metadata.json file to description of spatial outputs


allyhawkins:
Since we are now including the `SCPCL000000_metadata.json` file in the download for a spatial library and it will be included in the spatial folder for each library, it will need to be added to the list of files that are expected in the download listed in the [Spatial transcriptomics section](https://github.com/AlexsLemonade/scpca-docs/blob/main/docs/download_files.md#spatial-transcriptomics-libraries).

jaclyn-taroni:
Closed via #71.

PR #71, opened 2022-05-04
Update spatial download information to indicate spatial files will be separate

### LINKED ISSUE #67: Update download files section to reflect that spatial libraries and single-cell will be separate downloads
Currently in the section on download files, we have a few sentences noting that if projects or samples contain spatial they will have additional folders with spatial related output files. Since we now have decided that spatial files and single-cell/bulk files will be separate downloads, we should make sure that it's clear in the description that they will not be included in the same download.

allyhawkins:
Closes #67 and #69. 

Here I updated the download files section to reflect that the spatial libraries and single-cell/nuclei data will be available separately. To do that I removed mention of the spatial data in the beginning sections and then included a sentence at the beginning of the spatial specific section that stated the download would only contain spatial output files. Is this enough to indicate that they would be separate or should we make notes elsewhere in the text? 

I also updated the contents of the spatial download to include the `SCPCL000000_metadata.json` file as we [had indicated in our prior meetings](https://docs.google.com/document/d/1z8XwF7u7byPluDXPXKfYz0j4IhFNNaWzRWCH1-yiICA/edit). And then mentioned the inclusion of the `spatial_metadata.tsv` file which will be replacing the `single_cell_metadata.tsv` file.

### DIFF
--- a/docs/download_files.md
+++ b/docs/download_files.md
@@ -14,9 +14,6 @@ The files associated with each library are (example shown for a library with ID 
 
 Every download also includes a single `single_cell_metadata.tsv` file containing metadata for all libraries included in the download.
 
-If a sample includes a library that was processed using spatial transcriptomics, there will be an additional folder in the sample directory, `SCPCL000000_spatial`. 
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
Looks good; just one little comment.

Issue #76, opened 2022-05-09
FAQ: Why do multiplexed samples have est. demux counts?


jaclyn-taroni:
Should follow "What are est. demux cell counts?" (#73) and cover why we don't separate these samples out for folks currently!

jaclyn-taroni:
Closed via https://github.com/AlexsLemonade/scpca-docs/pull/83

Issue #73, opened 2022-05-09
FAQ: What are est. demux cell counts?


jaclyn-taroni:
We'll be adding a column to the metadata (and the sample table on the frontend) that contains estimated demultiplexed cell counts associated with a sample (`Est. Demux Cell Counts`). We should include an FAQ that explains what this field is.

jaclyn-taroni:
Closed via #83

PR #70, opened 2022-04-26
Add multiplex file docs

jashapiro:
This PR adds docs to describe the processing and file contents for multiplexed libraries. 

I struggled a lot to decide what the level of detail should be, so I welcome any thoughts about where there should be more or less detail. For example, I only described the "final" calls for the various demultiplexing programs, and left the (many) individual statistics descriptions as an "exercise to the reader". Let me know if you think I shoudl take a different approach there.

I also tried to update the description of the folder structure and library/sample relationships... I don't love anything that I wrote, but I think it gets the idea across? Let me know if you have improvements.

The folder naming may be in flux, I suppose. I currently say that multiple samples are comma separated, as that is the current output structure, but that seems like it could still change when building the download files.

I also think we decided that that library metadata file would have multiple lines per library, so I didn't think there was any need to change the description of that table.

I made a couple other changes, like including the folder images in the repo. I think this will should (it does for local builds), but if there is some reason those were not included before, please let me know!

Closes #64, but I am leaving it as a draft for now.

PR #83, opened 2022-05-31
Addition of Multiplex FAQs

### LINKED ISSUE #77: FAQ: What is a multiplexed sample?
Popping in some language discussed during today's meeting as a starting point:

> This project contains multiplexed libraries; multiple samples are included in a single library. These libraries have not been separated by samples (i.e., demultiplexed).

allyhawkins:
Closes #77, #73, and #76. 
Stacked on #70. 

I'm adding the three FAQs for multiplexed samples here, what is a multiplexed sample, what are est. demux cell counts, and why do multiplexed samples have est. demux cell counts. I decided to pull this together into one PR, because I thought it would make sense for reviewers to look at all of these together as they are all somewhat linked, especially the ones about estimated demux cell counts. 

I also stacked this on #70 so that I could link to some of the information in the processing and sce file contents docs in answering the questions. 

The one main question I have is if we want to explicitly say that we are reporting the cell counts using genetic demultiplexing? Right now as things are that's what we are using, but I know there was a concern that in the future that wouldn't always be the case. But it feels like something we probably want to include? 

Additionally, is there anymore detail or information that I'm missing that should be included in these explanations? 

As discussed in sprint planning I will request @jashapiro for a first round of review and then request a second reviewer who is less familiar to get their input.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -66,10 +66,11 @@ Downloading a multiplexed sample on the portal will result in obtaining the gene
 This means that users will receive the gene expression data for all samples that were combined into a given library and will have to separate any cells corresponding to the sample of interest before proceeding with downstream analysis.
 
 We have applied multiple {ref}`demultiplexing methods <processing_information:hto demultiplexing>` to multiplexed libraries and noticed that these demultiplexing methods can vary both in calls and confidence levels assigned. 
+[Here we have performed some exploratory analysis comparing demultiplexing methods within a single multiplexed library](https://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/main/analysis/quantifier-comparisons/15-demux-comparisons.nb.html).
 Because of the inconsistency across demultiplexing methods used, the choice of demultiplexing method to use is up to the discretion of the user. 
 Rather than separating out each sample, the sample calls and any associated statistics regarding sample calls for multiple demultiplexing methods can be found in the `_filtered.rds` file for each multiplexed library. 
-We also include the Hash Tag Oligo counts matrix to allow demultiplexing using other available methods.
 See the {ref}`demultiplexing results section <sce_file_contents:demultiplexing results>` for instructions on how to access the demultiplexing results in the `SingleCellExperiment` objects for multiplexed libraries.
+We also include the Hash Tag Oligo counts matrix to allow demultiplexing using other available methods.
 
 ## What are estimated demux cell counts? 
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -60,20 +60,25 @@ This means that a single library contains cells or nuclei that correspond to mul
 Each sample has been tagged with a hashtag oligo (HTO) prior to mixing, and that HTO can be used to identify which cells or nuclei belong to which sample within a multiplexed library. 
 The libraries available for download on the portal have not been separated by sample (i.e. demultiplexed), and therefore contain data from multiple samples.  
 
+## Why are demultiplexed samples not available? 
+
+Downloading a multiplexed sample on the portal will result in obtaining the gene expression files corresponding to the library containing the chosen multiplexed sample and any other samples that were multiplexed with that chosen sample. 
+This means that users will receive the gene expression data for all samples that were combined into a given library and will have to separate any cells corresponding to the sample of interest before proceeding with downstream analysis.
+
+We have applied multiple {ref}`demultiplexing methods <processing_information:hto demultiplexing>` to multiplexed libraries and noticed that these demultiplexing methods assign different cell barcodes to each sample at varying confidence levels. 
+Because of the inconsistency across demultiplexing methods used, the choice of demultiplexing method to use is up to the discretion of the user. 
+Rather than separating out each sample, the sample calls and any associated statistics regarding sample calls for multiple demultiplexing methods can be found in the `_filtered.rds` file for each multiplexed library. 
+See the {ref}`demultiplexing results section <sce_file_contents:demultiplexing results>` for instructions on how to access the demultiplexing results in the `SingleCellExperiment` objects for multiplexed libraries.
+
 ## What are estimated demux cell counts? 
 
 Estimated demux cell counts are provided for multiplexed libraries and refer to the estimated cell counts for each sample that is present in the library. 
-Cells that are included in the estimated cell count are cells that can be confidently assigned to a given sample using a given demultiplexing method. 
+In order to provide an estimate of the number of cells or nuclei that are present in a given sample before download, we use the estimated number of cells per sample identified by one of the tested demultiplexing methods.
+However, these estimated demux cell counts simply represent an estimate and we encourage users to investigate the data on their own and make their own decisions on the best demultiplexing method to use for their research purposes. 
+
 Note that not all cells in a library are included in the estimated demux cell count, as some cells may not be able to be assigned to a particular sample. 
-For more about demultiplexing, see the section on {ref}`processing multiplexed libraries <processing_information:multiplexed libraries>`.
-
-## Why do multiplexed samples have estimated demux cell counts? 
-
 Estimated demux cell counts will only be reported for multiplexed samples and will not be reported for single-cell or single-nuclei samples that are not multiplexed. 
-This is because multiplexed samples have multiple samples combined into one library using cell hashing, and samples have not been separated or demultiplexed. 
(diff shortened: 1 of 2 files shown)


jashapiro (commented):
Thank you for writing these up! I think my main comments are related to the last two questions listed, which seem largely related (though subtly different). I think we might see how combining them into a single answer works, as I expect people to read both together, so having redundant information seems wasteful. I do think the more direct "Why no demuxed samples" question should probably be added, which may allow some separation of some of the points embedded in the answers to the other two questions, but not explicitly stated.

allyhawkins:
Thanks for taking a look at this @jashapiro! I would agree that the second two questions seemed a little repetitive and inter-related. What I did was combine them into one answer, focusing on how we are providing an estimate of cell counts before download but they are simply an estimate. Let me know if the updated answer has a better answer? Or if there was something else you had in mind? I tried to make the answer more focused. 

I also added a new question for "Why are demultiplexed samples not available?" and explained that there is little consistency between the results of demultiplexing methods and that users can find the results in the SCE object and should make decisions on their own. I also reordered them so that this question appears first and then you have the question about what are estimated demux cell counts last. I thought this made sense since a lot of the reasoning behind estimated demux counts is because we aren't separating samples. I could repeat some of the same rationale about demultiplexing giving differing results in the "what are estimated demux cell counts" if we think that having it in both FAQs would be important. This should be ready for another look.

docs/faq.md
```diff
@@ -53,6 +53,33 @@ However, for most participants, only a single sample was collected and submitted
...
+Each sample has been tagged with a hashtag oligo (HTO) prior to mixing, and that HTO can be used to identify which cells or nuclei belong to which sample within a multiplexed library. 
+The libraries available for download on the portal have not been separated by sample (i.e. demultiplexed), and therefore contain data from multiple samples.  
+
+## Why are demultiplexed samples not available? 
+
+Downloading a multiplexed sample on the portal will result in obtaining the gene expression files corresponding to the library containing the chosen multiplexed sample and any other samples that were multiplexed with that chosen sample. 
+This means that users will receive the gene expression data for all samples that were combined into a given library and will have to separate any cells corresponding to the sample of interest before proceeding with downstream analysis.
+
+We have applied multiple {ref}`demultiplexing methods <processing_information:hto demultiplexing>` to multiplexed libraries and noticed that these demultiplexing methods assign different cell barcodes to each sample at varying confidence levels.
```
jashapiro, line 68: I wanted to include genetic demux in here as well, so I removed the reference cell barcodes and made a more general statement.

Do we want to link to the notebook where we explore this at all?
```suggestion
We have applied multiple {ref}`demultiplexing methods <processing_information:hto demultiplexing>` to multiplexed libraries and noticed that these demultiplexing methods can vary both in calls and confidence levels assigned. 
```

docs/faq.md
```diff
@@ -53,6 +53,33 @@ However, for most participants, only a single sample was collected and submitted
...
+
+## Why are demultiplexed samples not available? 
+
+Downloading a multiplexed sample on the portal will result in obtaining the gene expression files corresponding to the library containing the chosen multiplexed sample and any other samples that were multiplexed with that chosen sample. 
+This means that users will receive the gene expression data for all samples that were combined into a given library and will have to separate any cells corresponding to the sample of interest before proceeding with downstream analysis.
+
+We have applied multiple {ref}`demultiplexing methods <processing_information:hto demultiplexing>` to multiplexed libraries and noticed that these demultiplexing methods assign different cell barcodes to each sample at varying confidence levels. 
+Because of the inconsistency across demultiplexing methods used, the choice of demultiplexing method to use is up to the discretion of the user. 
+Rather than separating out each sample, the sample calls and any associated statistics regarding sample calls for multiple demultiplexing methods can be found in the `_filtered.rds` file for each multiplexed library.
```
jashapiro, line 70: I wanted to add a bit about the raw data here. We might want this sentence at the end though, not before the link to the demux results section?
```suggestion
Rather than separating out each sample, the sample calls and any associated statistics regarding sample calls for multiple demultiplexing methods can be found in the `_filtered.rds` file for each multiplexed library. 
We also include the Hash Tag Oligo counts matrix to allow demultiplexing using other available methods.
```

docs/faq.md
```diff
@@ -53,6 +53,33 @@ However, for most participants, only a single sample was collected and submitted
...
+Rather than separating out each sample, the sample calls and any associated statistics regarding sample calls for multiple demultiplexing methods can be found in the `_filtered.rds` file for each multiplexed library. 
+See the {ref}`demultiplexing results section <sce_file_contents:demultiplexing results>` for instructions on how to access the demultiplexing results in the `SingleCellExperiment` objects for multiplexed libraries.
+
+## What are estimated demux cell counts? 
+
+Estimated demux cell counts are provided for multiplexed libraries and refer to the estimated cell counts for each sample that is present in the library. 
+In order to provide an estimate of the number of cells or nuclei that are present in a given sample before download, we use the estimated number of cells per sample identified by one of the tested demultiplexing methods.
+However, these estimated demux cell counts simply represent an estimate and we encourage users to investigate the data on their own and make their own decisions on the best demultiplexing method to use for their research purposes.
```
jashapiro, line 77: Slight rewording
```suggestion
However, these estimated demux cell counts should only be considered a guide; we encourage users to investigate the data on their own and make their own decisions on the best demultiplexing method to use for their research purposes. 
```

docs/faq.md
```diff
@@ -53,6 +53,33 @@ However, for most participants, only a single sample was collected and submitted
...
+See the {ref}`demultiplexing results section <sce_file_contents:demultiplexing results>` for instructions on how to access the demultiplexing results in the `SingleCellExperiment` objects for multiplexed libraries.
+
+## What are estimated demux cell counts? 
+
+Estimated demux cell counts are provided for multiplexed libraries and refer to the estimated cell counts for each sample that is present in the library. 
+In order to provide an estimate of the number of cells or nuclei that are present in a given sample before download, we use the estimated number of cells per sample identified by one of the tested demultiplexing methods.
+However, these estimated demux cell counts simply represent an estimate and we encourage users to investigate the data on their own and make their own decisions on the best demultiplexing method to use for their research purposes. 
+
+Note that not all cells in a library are included in the estimated demux cell count, as some cells may not be able to be assigned to a particular sample.
```
jashapiro, line 79:
```suggestion
Note that not all cells in a library are included in the estimated demux cell count, as some cells may not have been assigned to a sample. 
```

docs/faq.md
```diff
@@ -53,6 +53,33 @@ However, for most participants, only a single sample was collected and submitted
...
+See the {ref}`demultiplexing results section <sce_file_contents:demultiplexing results>` for instructions on how to access the demultiplexing results in the `SingleCellExperiment` objects for multiplexed libraries.
+
+## What are estimated demux cell counts? 
+
+Estimated demux cell counts are provided for multiplexed libraries and refer to the estimated cell counts for each sample that is present in the library. 
+In order to provide an estimate of the number of cells or nuclei that are present in a given sample before download, we use the estimated number of cells per sample identified by one of the tested demultiplexing methods.
+However, these estimated demux cell counts simply represent an estimate and we encourage users to investigate the data on their own and make their own decisions on the best demultiplexing method to use for their research purposes. 
+
+Note that not all cells in a library are included in the estimated demux cell count, as some cells may not be able to be assigned to a particular sample. 
+Estimated demux cell counts will only be reported for multiplexed samples and will not be reported for single-cell or single-nuclei samples that are not multiplexed.
```
jashapiro, line 80: Matching tense to the other sentences
```suggestion
Estimated demux cell counts are only reported for multiplexed samples and are not reported for single-cell or single-nuclei samples that are not multiplexed. 
```

jashapiro (commented):
I like this organization much better! I think you have the order of the two sentences correct as well. 

I had a few remaining comments and a question about whether we should link to the notebook in https://github.com/AlexsLemonade/alsf-scpca/blob/main/analysis/quantifier-comparisons/15-demux-comparisons.nb.html (Linking to the html file there isn't great, but it is something).

allyhawkins:
@jashapiro I went ahead and added in a link to the notebook for now. I included a separate sentence with the link stating that we had performed some exploratory analysis comparing the different methods. Let me know what you think of that. 

Alternatively, and probably outside the scope of this PR, we could do something similar to what we did with the FAQ comparing alevin-fry to Cell Ranger and create specific figures and link to those rather than the html file, if there's something in particular that you think would be helpful rather than the entire notebook. But I leave that discretion up to you.

jashapiro (approved):
This looks good to me! I can't quite think of a good figure to include, so I think the link to the notebook is probably the best option for now.

PR #91, opened 2022-08-24
Add link to downstream analyses repo 

### LINKED ISSUE #88: Link to scpca-downstream-analyses in getting started section
After we make the `scpca-downstream-analyses` repo public and have the core module ready, we should include a link in the portal docs so that users that download data from the portal can navigate to `scpca-downstream-analyses` if they want to use it for analyzing the data they have downloaded. To me this makes sense to live in the `Getting Started with an ScPCA dataset` section of the docs.

allyhawkins:
Closes #88 

Here I added a link to the downstream analyses repo in the getting started section. I wasn't quite sure how much detail we wanted to put here, but I wanted to emphasize that there is a workflow that processes samples in parallel and performs the steps mentioned in the getting started doc. I think this is probably the only place in the docs where it makes sense to include a link. 

Also there's some trimming of white space, so apologies for the extra lines in the diff.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -2,6 +2,9 @@
 
 This section provides information on next steps you might take after downloading a dataset from the ScPCA portal.
 
+More resources for processing single-cell datasets can be found on our Github in the [`scpca-downstream-analyses` repository](https://github.com/AlexsLemonade/scpca-downstream-analyses).
+This repository houses a workflow that can be used to perform initial filtering, dimensionality reduction, normalization, and clustering on single-cell and single-nuclei samples in parallel.
+
 ## Importing ScPCA data into R
 
 Quantified single-cell or single-nuclei gene expression data is provided as an RDS file as described in the {ref}`single cell gene expression file contents section<sce_file_contents:single-cell gene expression file contents>`.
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -2,6 +2,9 @@
 
 This section provides information on next steps you might take after downloading a dataset from the ScPCA portal.
 
+We have also provided a workflow that can be used to perform initial filtering, dimensionality reduction, normalization, and clustering on single-cell and single-nuclei samples in parallel.
+This workflow and more resources for processing single-cell datasets can be found on our Github in the [`scpca-downstream-analyses` repository](https://github.com/AlexsLemonade/scpca-downstream-analyses).
```
cbethell, line 6:
```suggestion
This workflow and more resources for processing single-cell/single-nuclei datasets can be found on our Github in the [`scpca-downstream-analyses` repository](https://github.com/AlexsLemonade/scpca-downstream-analyses).
```

cbethell: Or something like the above suggestion since you do mention both types of samples in the first sentence there?

cbethell (approved):
Thanks for doing this @allyhawkins! It looks good to me -- I think the level of detail here is suffice, I just left one comment below but I don't think there will be a need for me to re-review 👍

PR #108, opened 2023-05-09
Document changes to participant ID requirements

### LINKED ISSUE #107: Tweak FAQ and downloaded files section for participant IDs to allow for future additions
There will be projects released shortly that do not currently have multiple samples that map to the same participant, but we expect future releases to include samples to meet that condition. The investigator has asked us to include participant IDs so that we don't have to make any revisions later.

As such, we're going to need to change the language in this FAQ to cover these cases: https://github.com/AlexsLemonade/scpca-docs/blob/5c66caa36f99b764f9f770e282b5787e318f3029/docs/faq.md#why-do-some-samples-have-missing-participant-ids

**Edit:** Also need to address this section: https://github.com/AlexsLemonade/scpca-docs/blob/5c66caa36f99b764f9f770e282b5787e318f3029/docs/download_files.md#metadata, so I've updated the title accordingly.

Anyone who implements this should find all "participant" to be sure everything has been addressed.

allyhawkins:
Closes #107 

I updated the metadata table in the download files section to mention that the participant ID is required for all participants that have multiple samples, but optional in other cases. I also updated the FAQ to reflect when the ID is required vs. optional.

### DIFF
--- a/docs/download_files.md
+++ b/docs/download_files.md
@@ -71,7 +71,7 @@ The `single_cell_metadata.tsv` file is a tab-separated table with one row per li
 | `technology`      | 10X kit used to process library                                |
 | `filtered_cell_count` | Number of cells after filtering with `emptyDrops`          |
 | `submitter_id`    | Original sample identifier from submitter                      |
-| `participant_id`  | Original participant id, if there are multiple samples from the same participant                                                                        |
+| `participant_id`  | Original participant id, required to be present if there are multiple samples from the same participant, optional for all other samples                                                                        |
 | `submitter`       | Submitter name/id                                              |
 | `age`             | Age at time sample was obtained                                |
 | `sex`             | Sex of patient that the sample was obtained from               |
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -50,7 +50,7 @@ The `participant_id`, when present, indicates the participant from which a colle
 For example, one participant may have a sample collected both at initial diagnosis and at relapse.
 This would result in two different sample ID's, but the same participant ID.
 However, for most participants, only a single sample was collected and submitted for sequencing.
-Because of this, many samples do not have a separate participant ID.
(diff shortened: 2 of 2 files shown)


docs/faq.md
```diff
@@ -51,7 +51,8 @@ For example, one participant may have a sample collected both at initial diagnos
 This would result in two different sample ID's, but the same participant ID.
 However, for most participants, only a single sample was collected and submitted for sequencing.
 Because of this, many of the samples do not have a separate participant ID.
-Participant IDs are only present for samples that were derived from the same participant as at least one other sample.
+A `participant_id` is required for all samples that were derived from the same participant as at least one other sample.
+They may also be present for samples in which there is only one sample collected from a participant, but for this scenario the presence of a `participant_id` is optional.
```
jaclyn-taroni, line 55:
```suggestion
They may also be present for samples in which only one sample is collected from a participant, but for this scenario, the presence of a `participant_id` is optional.
```

docs/download_files.md
```diff
@@ -71,7 +71,7 @@ The `single_cell_metadata.tsv` file is a tab-separated table with one row per li
 | `technology`      | 10X kit used to process library                                |
 | `filtered_cell_count` | Number of cells after filtering with `emptyDrops`          |
 | `submitter_id`    | Original sample identifier from submitter                      |
-| `participant_id`  | Original participant id, if there are multiple samples from the same participant                                                                        |
+| `participant_id`  | Original participant id, required to be present if there are multiple samples from the same participant, optional for all other samples                                                                        |
```
jaclyn-taroni, line 74:
```suggestion
| `participant_id`  | Original participant id, required when there are multiple samples from the same participant, optional for all other samples                                                                        |
```

jaclyn-taroni (approved):
Returning a few wordsmithing items to improve clarity. Approving with the expectation that you'll accept those 👍🏻

PR #121, opened 2023-06-28
Updates for ADT normalization

### LINKED ISSUE #120: Update ADT normalization docs
For CITE-seq, we're going to fall back on `log1p` if size factors fail rather than not normalizing at all. We'll need to document this!

Source: https://github.com/AlexsLemonade/scpca-nf/pull/343#discussion_r1245324370

sjspielman:
Closes #120 

Two main updates:
- There is an `adt_normalization` field now in the metadata for CITE-seq samples
- We no longer fail if size factors are 0. Instead, `log1p`.

Let me know if what I've written is clear, thanks!

### DIFF
--- a/components/dictionary.txt
+++ b/components/dictionary.txt
@@ -48,6 +48,7 @@ Kaminow
 Marioni
 Pearson
 pre
+pseudocount
 pseudogenes
 Lun
 miQC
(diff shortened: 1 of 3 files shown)


allyhawkins (approved):
LGTM!

PR #134, opened 2023-09-01
Add documentation for celltype annotation

sjspielman:
Closes #130 

This PR adds docs for cell typing as described in #130, largely based on the comments I wrote after the issue was opened. 
Let me know where I can tweak any phrasing, what can use more or less detail, etc. Also, anything I am entirely missing?

PR #141, opened 2023-09-05
Update config file

sjspielman:
This PR hopefully fixes the config file, because: https://blog.readthedocs.com/migrate-configuration-v2/

First noticed in https://github.com/AlexsLemonade/scpca-docs/pull/134#issuecomment-1706915911


sjspielman:
Hm, this PR might be unnecessary. The build might be failing for other reasons that have nothing to do with needed to bump syntax; specifically, github status is a mess right now. The error message below could also be b/c I tried to see what happens if I move the `.readthedocs.yml` into the top level, but since it's all hanging...who knows. We'll wait for github to sort itself out!

Screenshot 2023-09-05 at 12 54 37 PM

sjspielman:
Well, seems to be working now! I think it actually was a combination of:

- The config _did_ need to move back to top-level directory, but also when I did that the first time, github was broken 😭 
- Moved where the python version was declared. Might not have been necessary, but matches the template config so I say we keep this syntax 
- I probably didn't need to remove `formats = []`, but does it matter that I did?

allyhawkins (approved):
Thanks for doing this! And I think removing `formats` is okay.

PR #153, opened 2023-09-27
Document AnnData in downloads

allyhawkins:
Closes #144 

This PR makes changes to the download section to document data available as AnnData. 

- In the intro describing the contents of the download, I included some rephrasing to mention that downloads include either SCE or AnnData objects/ files. I also did some rearranging and moved things mentioning the bulk data and spatial data to the main intro section rather than below the screenshots.
- I created separate headers for SCE downloads and AnnData downloads. Within each header, I have a header for project vs. sample download that will include the illustrations. 
- For AnnData, I included a separate header for downloading a sample with CITE-seq. 
- I added notes that multiplexed samples are only available as SCE objects.

PR #154, opened 2023-09-27
Include description of sample metadata present in SCE object 

### LINKED ISSUE #146: Update components of SCE section to include added metadata fields
We have recently added a bunch of new things that are now stored in the metadata of the SCE objects, like the `sample_metadata`. We will need to update the [`Components of a SingleCellExperiment object`](https://scpca.readthedocs.io/en/latest/sce_file_contents.html#components-of-a-singlecellexperiment-object) section to reflect these new changes. 

This includes: 
- the addition of `sample_metadata` 
- `assay_ontology_term_id` 
- `seq_unit`

allyhawkins:
Closes #146 
Stacked on #153 

This PR modifies the current SCE contents to include a table listing all of the sample metadata columns that are present in `metadata(sce)$sample_metadata`. I added a row for listing the `sample_metadata` in the table with the `metadata(sce)` contents, but then I also made a separate section just for the sample metadata. Here, I added a table showing the expected columns in the sample metadata table that's stored in the object. I also added a note that there may be additional columns depending on the project. 

I added in the `seq_unit` and `assay_ontology_term_id` to the metadata list since those had not been added yet. 

While working on this, I also removed the note about `participant_id` only being filled in for participants that contain multiple samples. Those are now present for all of our samples.

### DIFF
--- a/docs/download_files.md
+++ b/docs/download_files.md
@@ -87,7 +87,7 @@ The `single_cell_metadata.tsv` file is a tab-separated table with one row per li
 | `technology`      | 10x kit used to process library                                |
 | `filtered_cell_count` | Number of cells after filtering with `emptyDrops`          |
 | `submitter_id`    | Original sample identifier from submitter                      |
-| `participant_id`  | Original participant id, required when there are multiple samples from the same participant, optional for all other samples                                                                        |
+| `participant_id`  | Original participant id                                        |
 | `submitter`       | Submitter name/id                                              |
 | `age`             | Age at time sample was obtained                                |
 | `sex`             | Sex of patient that the sample was obtained from               |
--- a/docs/sce_file_contents.md
+++ b/docs/sce_file_contents.md
@@ -106,6 +106,8 @@ expt_metadata <- metadata(sce)
 | `usa_mode`          | Boolean indicating whether quantification was done using `alevin-fry` USA mode                                                 |
 | `af_num_cells`      | Number of cells reported by `alevin-fry`                                                                                       |
 | `tech_version`      | A string indicating the technology and version used for the single-cell library, such as 10Xv2, 10Xv3, or 10Xv3.1              |
+| `assay_ontology_term_id` | A string indicating the [Experimental Factor Ontology](https://www.ebi.ac.uk/ols/ontologies/efo) term id associated with the `tech_version`  |
+| `seq_unit`         | `cell` for single-cell samples or `nucleus` for single-nucleus samples                                                          |
 | `transcript_type`   | Transcripts included in gene counts: `spliced` for single-cell samples and `unspliced` for single-nuclei                       |
 | `miQC_model`        | The model object that `miQC` fit to the data and was used to calculate `prob_compromised`. Only present for `filtered` objects |
 | `filtering_method`  | The method used for cell filtering. One of `emptyDrops`, `emptyDropsCellRanger`, or `UMI cutoff`. Only present for `filtered` objects |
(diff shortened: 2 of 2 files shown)


docs/sce_file_contents.md
```diff
@@ -127,6 +130,36 @@ expt_metadata <- metadata(sce)
...
+
+Relevant sample metadata is available as a data frame stored in the `metadata(sce)$sample_metadata` slot of the `SingleCellExperiment` object.
+Each row in the data frame will correspond to a sample present in the library.
+The following columns are included in the sample metadata data frame for all libraries.
+
+| Column name   | Contents                                                         |
+| ------------- | ---------------------------------------------------------------- |
+| `sample_id`   | Sample ID in the form `SCPCS000000`                            |
+| `library_id`   | Library ID in the form `SCPCL000000`                             |
+| `donor_id`  | Unique id corresponding to the donor from which the sample was obtained (equivalent to `particpant_id` included in `single_cell_metadata.tsv`) |
```
jashapiro, line 143: Is this naming difference because of CELLxGENE?

docs/sce_file_contents.md
```diff
@@ -127,6 +130,36 @@ expt_metadata <- metadata(sce)
...
+| `submitter`       | Submitter name/id                                              |
+| `age`             | Age at time sample was obtained                                |
+| `sex`             | Sex of patient that the sample was obtained from               |
+| `diagnosis`       | Tumor type                                                     |
+| `subdiagnosis`    | Subcategory of diagnosis or mutation status (if applicable)    |
+| `tissue_location` | Where in the body the tumor sample was located                 |
+| `disease_timing`  | At what stage of disease the sample was obtained, either diagnosis or recurrence |
+| `organism`         | The organism the sample was obtained from (e.g., `Homo_sapiens`) |
+| `development_stage_ontology_term_id` | `HsapDv` ontology term indicating developmental stage - for _Homo sapiens_ either one of the terms for [1 - 11 months old](https://ontobee.org/search?ontology=HSAPDV&keywords=month-old&submit=Search+terms) or for [older than 12 months old](https://ontobee.org/search?ontology=HSAPDV&keywords=year-old&submit=Search+terms). If unavailable, `unknown` is used.  |
```
jashapiro, line 153: Do we need this level of detail? I would maybe just link to http://obofoundry.org/ontology/hsapdv.html or https://www.ebi.ac.uk/ols/ontologies/hsapdv. The former is probably better for consistency, even if I like EBI's interface a bit better

For our internal use when creating these, we might need the instruction about which terms to use, but when people are looking up the terms, they should not need quite the same level of detail?

```suggestion
| `development_stage_ontology_term_id` | [`HsapDv` ontology](http://obofoundry.org/ontology/hsapdv.html) term indicating developmental stage. If unavailable, `unknown` is used.  |
```

docs/sce_file_contents.md
```diff
@@ -127,6 +130,36 @@ expt_metadata <- metadata(sce)
...
+| `sex_ontology_term_id` | [PATO](http://obofoundry.org/ontology/pato.html) term referring to the sex of the sample, must be a child of [`PATO:0001894`](https://ontobee.org/ontology/PATO?iri=http://purl.obolibrary.org/obo/PATO_0001894). If unavailable, `unknown` is used.  |
```
jashapiro, line 154: Here again, I'm not sure we need the info about where in the ontology this is required to be, as that is more about upstream creation than downstream use.

If we are going to link to specific terms, we should probably use purl links?
i.e., http://purl.obolibrary.org/obo/PATO_0001894


```suggestion
 `sex_ontology_term_id` | [PATO](http://obofoundry.org/ontology/pato.html) term referring to the sex of the sample. If unavailable, `unknown` is used.  |
```

docs/sce_file_contents.md
```diff
@@ -127,6 +130,36 @@ expt_metadata <- metadata(sce)
...
+| `organism_ontology_id` | NCBI taxonomy term for organism, e.g. [`NCBITaxon:9606`](https://ontobee.org/ontology/NCBITaxon?iri=http://purl.obolibrary.org/obo/NCBITaxon_9606). |
```
jashapiro, line 155: I don't know if we need the NCBI link or if we should link to obofoundry as we do elsewhere, but again, purl URL
```suggestion
| `organism_ontology_id` | [NCBI taxonomy](https://www.ncbi.nlm.nih.gov/taxonomy) term for organism, e.g. [`NCBITaxon:9606`](http://purl.obolibrary.org/obo/NCBITaxon_9606). |
```

docs/sce_file_contents.md
```diff
@@ -106,7 +106,10 @@ expt_metadata <- metadata(sce)
 | `usa_mode`          | Boolean indicating whether quantification was done using `alevin-fry` USA mode                                                 |
 | `af_num_cells`      | Number of cells reported by `alevin-fry`                                                                                       |
 | `tech_version`      | A string indicating the technology and version used for the single-cell library, such as 10Xv2, 10Xv3, or 10Xv3.1              |
+| `assay_ontology_term_id` | A string indicating the [Experimental Factor Ontology](https://www.ebi.ac.uk/ols/ontologies/efo) term id associated with the `tech_version`  |
```
jashapiro, line 109: I was going to comment here to suggest that we link to obofoundry for consistency, only to find that doesn't have EFO. So this is fine. 

(I was then going to suggest linking directly to OLS4 to avoid warnings, but I see that this is scheduled to be replaced at the end of next month, so we are probably fine)

docs/sce_file_contents.md
```diff
@@ -127,6 +130,36 @@ expt_metadata <- metadata(sce)
...
+| `sex_ontology_term_id` | [PATO](http://obofoundry.org/ontology/pato.html) term referring to the sex of the sample, must be a child of [`PATO:0001894`](https://ontobee.org/ontology/PATO?iri=http://purl.obolibrary.org/obo/PATO_0001894). If unavailable, `unknown` is used.  |
+| `organism_ontology_id` | NCBI taxonomy term for organism, e.g. [`NCBITaxon:9606`](https://ontobee.org/ontology/NCBITaxon?iri=http://purl.obolibrary.org/obo/NCBITaxon_9606). |
+| `self_reported_ethnicity_ontology_term_id` | For _Homo sapiens_, a [`Hancestro` term](http://obofoundry.org/ontology/hancestro.html). `multiethnic` indicates more than one ethnicity is reported. `unknown` indicates unavailable ethnicity and `NA` is used for all other organisms.  |
+| `disease_ontology_term_id` | [`MONDO`](http://obofoundry.org/ontology/mondo.html) term indicating disease type. [`PATO:0000461`](https://ontobee.org/ontology/PATO?iri=http://purl.obolibrary.org/obo/PATO_0000461) indicates normal or healthy tissue. If unavailable, `NA` is used.  |
```
jashapiro, line 157:
```suggestion
| `disease_ontology_term_id` | [`MONDO`](http://obofoundry.org/ontology/mondo.html) term indicating disease type. [`PATO:0000461`](http://purl.obolibrary.org/obo/PATO_0000461) indicates normal or healthy tissue. If unavailable, `NA` is used.  |
```

jashapiro (commented):
This looks good overall. My main comments are about the level of detail in some of the fields and link consistency.

docs/sce_file_contents.md
allyhawkins: Yea... When I was writing this I was thinking about whether or not we should not do this and only if we end up submitting to cellxgene.

allyhawkins:
@jashapiro thanks for catching those link inconsistencies! I believe I got all of them, but let me know if I missed one. I think the only thing left is to decide if we are keeping the `donor_id` or changing back to `participant_id`.

docs/sce_file_contents.md
jashapiro: We have to do other reformatting if we submit to CELLxGENE, right? If so, I would opt for consistency in the metadata and changing it just for submission. (I actually may like `donor_id` better, but it is probably harder to change that everywhere than to leave it!)

allyhawkins: The only thing we may have to change is the title from the project ids currently stored in the object to the actual title. We also may have to change other things depending on the schema version that is around when/ if we submit. I think I'm good with keeping `participant_id` throughout and then changing it before submission if needed.

docs/download_files.md
```diff
@@ -87,7 +87,7 @@ The `single_cell_metadata.tsv` file is a tab-separated table with one row per li
 | `technology`      | 10x kit used to process library                                |
 | `filtered_cell_count` | Number of cells after filtering with `emptyDrops`          |
 | `submitter_id`    | Original sample identifier from submitter                      |
-| `participant_id`  | Original participant id, required when there are multiple samples from the same participant, optional for all other samples                                                                        |
+| `participant_id`  | Original participant id                                        |
```
jashapiro, line 90: I feel like we should use the same language in both tables?
```suggestion
| `participant_id`  | Unique id corresponding to the donor from which the sample was obtained |
```

jashapiro (approved):
I didn't check for full harmonization between the shared columns of the two tables, but adjusted one...

Otherwise this should be good.

Issue #165, opened 2023-10-19
Add cell type annotation to getting started docs


allyhawkins:
Right now the getting started docs include information on how to access and work with the processed information we have in our objects (like PCA, UMAP, clusters, etc). We should include a section on cell type annotation. This should include both how to pull out the various cell types, how to find the associated references, and maybe some useful links on cell type annotation with SingleR/CellAssign. We probably also want to mention the supplemental cell type report and what type of information you can find there.

sjspielman:
When requesting review here _from Jackie_, send her an object to test with!

sjspielman:
While we're here, also double back to `sce_file_contents.md` to add these metadata fields:

https://github.com/AlexsLemonade/scpca-nf/issues/470#issuecomment-1802151588

sjspielman:
> When requesting review here _from Jackie_, send her an object to test with!

**WOOPS.** 

@jaclyn-taroni, let's chat :grimacing:! The updated getting started instructions have been reviewed and merged, but we can go around again for you to have a look separately as someone not too in the weeds of developing the code. Since the PR where changes were made is closed, I could maybe revert the merge and open a new PR to go again? Do you think that is reasonable here? 

Here's the current version of the file: https://github.com/AlexsLemonade/scpca-docs/blob/325bfdbf150babcd360510a891d5ec786a68708d/docs/getting_started.md

Let me know what you think for the right next steps and I will Slack you some files!! Sorry about missing that!

jaclyn-taroni:
It seems pretty reasonable. I will pick a nit, though – some of the publication references in the cell annotation section use `[]` instead of `()` and I don't know why.

sjspielman:
Closed by #193 and #204

Issue #159, opened 2023-10-12
Update download section to include supplemental cell type file 


allyhawkins:
We are adding an additional html report for all samples that have cell type annotation included. We will need to update the download section of the docs to mention that file. Additionally, the new html report must be added to the existing illustrations. 

It's my understanding that this file will be available with all libraries, so we should update the docs with that assumption.

allyhawkins:
Closed by #191

PR #166, opened 2023-10-19
Update SCE contents and processing with cell type information

### LINKED ISSUE #160: Update SCE contents section to include cell type annotation
Once we finalize how exactly we are storing the cell type annotations, we will want to ensure the docs describing the SCE contents (and AnnData) are up to date. This should include: 
- where the annotations live in the SCE object (`colData`) and what the column names are
- where the full results from cell typing live (`metadata`)
- where the information for cell typing methods and references used live (`metadata`)

allyhawkins:
Closes #160 

This PR makes some small adjustments to docs to sync up what we say about cell typing here to where the cell types actually live in the object. 

- I added a link for both SingleR and CellAssign papers to the processing information section on cell type annotation. 
- In that same section, I added a note about submitter annotations. 
- I included the `submitter_annotation` column in the cell metrics section. When adding this in, I chose to add this in the table that mentions items that are in both filtered and processed objects. I think this is the right place since if they are available they will be in all of the objects. 
- I standardized how we refer to the `cellassign_reference` and `singler_reference`. This matches changes made in https://github.com/AlexsLemonade/scpca-nf/pull/510 and https://github.com/AlexsLemonade/scpca-nf/pull/503. 

Please let me know if you think I missed any other areas here. I will update the download section separately as part of #159 and the getting started section as part of #165.

### DIFF
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -87,7 +87,7 @@ During annotation, we additionally include an `"other"` cell type that does not 
 As a consequence, cells which `CellAssign` cannot confidently annotate from the full marker gene list are labeled as `"other"`.
 
 **Note:** For some libraries, cell type annotations were provided from the group that submitted the original data.
-If this is the case, the cell type annotations obtained from the submitter will be present in addition to cell type annotation performed with `SingleR` and `CellAssign`.
+In these cases, the cell type annotations obtained from the submitter will be present in addition to cell type annotation performed with `SingleR` and `CellAssign`.
 
 ## ADT quantification from CITE-seq experiments
 
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -75,8 +75,8 @@ Finally, these principal components are used to calculate the [UMAP (Uniform Man
 
 We perform cell type annotation with two complementary methods, where possible:
 
-- [`SingleR`](https://bioconductor.org/packages/release/bioc/html/SingleR.html), a reference-based cell type annotation method
-- [`CellAssign`](https://github.com/Irrationone/cellassign), a marker-gene--based cell type annotation method
+- [`SingleR`](https://bioconductor.org/packages/release/bioc/html/SingleR.html), a reference-based cell type annotation method ([Looney _et al._ 2019](https://doi.org/10.1038/s41590-018-0276-y))
+- [`CellAssign`](https://github.com/Irrationone/cellassign), a marker-gene--based cell type annotation method ([Zhang _et al._ 2019](https://doi.org/10.1038/s41592-019-0529-1))
 
 For `SingleR` annotation, we identify an appropriate reference dataset from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html) and train the classification model to use ontology IDs for annotation.
 Cells which `SingleR` cannot confidently assign are labeled as `NA`.
@@ -85,6 +85,7 @@ For `CellAssign` annotation, we identify an appropriate set of marker genes for 
 We combine these marker genes with all "immune cell" `PanglaoDB` marker genes to create a full marker gene list for for cell type annotation.
 During annotation, we additionally include an `"other"` cell type that does not express any of these marker genes.
 As a consequence, cells which `CellAssign` cannot confidently annotate from the full marker gene list are labeled as `"other"`.
+
 ## ADT quantification from CITE-seq experiments
 
 CITE-seq libraries with reads from antibody-derived tags (ADTs) were also quantified using  [`salmon alevin`](https://salmon.readthedocs.io/en/latest/alevin.html) and [`alevin-fry`](https://alevin-fry.readthedocs.io/en/latest/), rounded to integer values.
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -86,6 +86,9 @@ We combine these marker genes with all "immune cell" `PanglaoDB` marker genes to
 During annotation, we additionally include an `"other"` cell type that does not express any of these marker genes.
 As a consequence, cells which `CellAssign` cannot confidently annotate from the full marker gene list are labeled as `"other"`.
 
+**Note:** For some libraries, cell type annotations were provided from the group that submitted the original data.
+If this is the case, the cell type annotations obtained from the submitter will be present in addition to cell type annotation performed with `SingleR` and `CellAssign`.
+
 ## ADT quantification from CITE-seq experiments
 
 CITE-seq libraries with reads from antibody-derived tags (ADTs) were also quantified using  [`salmon alevin`](https://salmon.readthedocs.io/en/latest/alevin.html) and [`alevin-fry`](https://alevin-fry.readthedocs.io/en/latest/), rounded to integer values.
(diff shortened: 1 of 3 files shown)


sjspielman:
@allyhawkins FYI I'm going to get to this today but realize I forgot it in my scrum!

docs/processing_information.md
```diff
@@ -85,6 +85,10 @@ For `CellAssign` annotation, we identify an appropriate set of marker genes for
 We combine these marker genes with all "immune cell" `PanglaoDB` marker genes to create a full marker gene list for for cell type annotation.
 During annotation, we additionally include an `"other"` cell type that does not express any of these marker genes.
 As a consequence, cells which `CellAssign` cannot confidently annotate from the full marker gene list are labeled as `"other"`.
+
+**Note:** For some libraries, cell type annotations were provided from the group that submitted the original data.
+If this is the case, the cell type annotations obtained from the submitter will be present in addition to cell type annotation performed with `SingleR` and `CellAssign`.
```
sjspielman, line 90:
```suggestion
In these cases, the cell type annotations obtained from the submitter will be present in addition to cell type annotation performed with `SingleR` and `CellAssign`.
```

docs/sce_file_contents.md
```diff
@@ -63,6 +63,8 @@ See the description of the {ref}`processed gene expression data <processing_info
 | `miQC_pass`             | Indicates whether the cell passed the default miQC filtering. `TRUE` is assigned to cells with a low probability of being compromised (`prob_compromised` < 0.75) or [sufficiently low mitochondrial content](https://bioconductor.org/packages/release/bioc/vignettes/miQC/inst/doc/miQC.html#preventing-exclusion-of-low-mito-cells)  |
 | `scpca_filter` | Labels cells as either `Keep` or `Remove` based on filtering criteria (`prob_compromised` < 0.75 and number of unique genes detected > 200) |
 | `adt_scpca_filter` | If CITE-seq was performed, labels cells as either `Keep` or `Remove` based on ADT filtering criteria (`discard = TRUE` as determined by [`DropletUtils::CleanTagCounts()`](https://rdrr.io/github/MarioniLab/DropletUtils/man/cleanTagCounts.html)) |
+| `submitter_annotation` | If available, cell type annotations obtained from the group that submitted the original data. Cells that are not classified by the submitter are labeled as "Unclassified cell" |
```
sjspielman, line 66: I think this is only true in the QC report, but we are working on that! In the SCE _itself_, these values are literal `NA`s.

```suggestion
| `submitter_celltype_annotation` | If available, cell type annotations obtained from the group that submitted the original data. Cells that the submitter did not annotate are labeled as `NA` |
```

docs/sce_file_contents.md
```diff
@@ -130,12 +132,11 @@ metadata(sce) # experiment metadata
 | `cluster_algorithm` | The algorithm used to perform graph-based clustering of cells. Only present for `processed` objects |
 | `cluster_weighting` | The weighting approach used during graph-based clustering. Only present for `processed` objects |
 | `cluster_nn`        | The nearest neighbor parameter value used for the graph-based clustering. Only present for `processed` objects |
-| `celltype_methods` | If cell type annotation was performed, a vector of the methods used for annotation. May include `"singler"` and/or `"cellassign"`. Only present for `processed` objects |
+| `celltype_methods` | If cell type annotation was performed, a vector of the methods used for annotation. May include `"submitter"`, `"singler"` and/or `"cellassign"`. If submitter cell-type annotations are available, this will be present in all objects, otherwise this item is only present for `processed` objects |
```
sjspielman, line 135:
```suggestion
| `celltype_methods` | If cell type annotation was performed, a vector of the methods used for annotation. May include `"submitter"`, `"singler"` and/or `"cellassign"`. If submitter cell-type annotations are available, this metadata item will be present in all objects. Otherwise, this item will only be in `processed` objects |
```

docs/sce_file_contents.md
```diff
@@ -130,12 +132,11 @@ metadata(sce) # experiment metadata
...
 | `singler_results` | If cell typing with `SingleR` was performed, the full result object returned by `SingleR` annotation. Only present for `processed` objects |
-| `singler_reference_name` | If cell typing with `SingleR` was performed, the name of the [`celldex`](http://bioconductor.org/packages/release/data/experiment/html/celldex.html) reference dataset used for annotation. Only present for `processed` objects |
-| `singler_reference_label` | If cell typing with `SingleR` was performed, the name of the label in the reference dataset used for annotation. Only present for `processed` objects |
```
sjspielman, line 136: I actually wonder if we should revive this and get it into the SCE metadata. It seems like useful information to have about celltyping, but also something to circle back to if we re-add it in. 
Let me know what you think about this!

----

If we do that, here's how we might go about it:

1. Modify [`classify_SingleR.R`](https://github.com/AlexsLemonade/scpca-nf/blob/development/bin/classify_SingleR.R) to also add the label name into the singler result metadata
2. Add lines here to add info to SCE metadata:
https://github.com/AlexsLemonade/scpca-nf/blob/934df294bbfe2a01d8ffdb94ba9f5d6b432a547d/bin/add_celltypes_to_sce.R#L100-L102
3. Bonus change: We could also change this line to check the label name directly instead of name presence, but either way should be fine. https://github.com/AlexsLemonade/scpca-nf/blob/934df294bbfe2a01d8ffdb94ba9f5d6b432a547d/bin/add_celltypes_to_sce.R#L75

sjspielman (commented):
Looks good overall! Left some small text comments, and one question about the SingleR reference label.

One other thing I noticed that isn't part of this PR but maybe we should do it - For the description of this column, we should state it's only present if ontology labels were used for annotation? Can rephrase to opening of sentence to: "If cell typing with `SingleR` was performed with ontology labels,...."

https://github.com/AlexsLemonade/scpca-docs/blob/9d9836393d639e0e43b145928c18053f6c465bc3/docs/sce_file_contents.md?plain=1#L76

docs/sce_file_contents.md
allyhawkins: I'm good with that. Was there a reason that we removed it in the first place?

allyhawkins:
Thanks for taking a look at this @sjspielman. I incorporated your comments and then also filed https://github.com/AlexsLemonade/scpca-nf/pull/522 to return the label name. I then updated the docs here to reflect that.

docs/sce_file_contents.md
sjspielman: > Was there a reason that we removed it in the first place?

No, I think it just got lost in the shuffle, especially when we changed the code to check for the ontology df in the model object instead of passing in the label parameter. Thanks for filing the PR for it (expect a review after YI, probably).

docs/sce_file_contents.md
```diff
@@ -71,7 +73,7 @@ Further, if cell type annotation was performed, there will be additional columns
 | ----------------------- | ----------------------------------------------------- |
 | `cluster`  | Cell cluster identity identified by graph-based clustering |
 | `singler_celltype_annotation`  | If cell typing with `SingleR` was performed, the annotated cell type. Cells labeled as `NA` are those which `SingleR` could not confidently annotate |
-| `singler_celltype_ontology`  | If cell typing with `SingleR` was performed, the annotated cell type's ontology ID. Cells labeled as `NA` are those which `SingleR` could not confidently annotate |
+| `singler_celltype_ontology`  | If cell typing with `SingleR` was performed with ontology labels, the annotated cell type's ontology ID. Cells labeled as `NA` are those which `SingleR` could not confidently annotate. |
```
sjspielman, line 76: Oh wait, no periods at the end; sorry if that was my bad!
```suggestion
| `singler_celltype_ontology`  | If cell typing with `SingleR` was performed with ontology labels, the annotated cell type's ontology ID. Cells labeled as `NA` are those which `SingleR` could not confidently annotate |
```

sjspielman (approved):
LGTM!

sjspielman:
for posterity noting that `singler_reference_label` metadata item referred to here matches the name used in https://github.com/AlexsLemonade/scpca-nf/pull/522 👍

PR #174, opened 2023-10-30
Update docs to note we always use spliced and unspliced cDNA

### LINKED ISSUE #142: Update processing to reflect we always use unspliced now
In https://github.com/AlexsLemonade/scpca-docs/blob/29f445099425fc25f1556387a3bf22af80053d9d/docs/processing_information.md?plain=1#L79-L85

we state that we are using different counts for nuclei and cells, but I don't think this is what we are actually doing anymore (since https://github.com/AlexsLemonade/scpca-nf/pull/242) or so, as we are trying to match Cell Ranger 7's new behavior.

allyhawkins:
Closes #142 
Note that this PR is getting merged into the new `documentation-improvements` branch that I made. This branch is protected and should be use for making all the pre-AnnData and pre-celltype changes. 

Here I updated the section on post processing alevin-fry to indicate that we always combine counts. I also added a sentence to the sce file contents to note that the `counts` matrix includes the counts from both spliced and unspliced cDNA and included a link to the processing information. 

I did include a sentence about the presence of the `spliced` assay in the object, but I'm unsure if we want to mention that or not... I'm curious about what others think.

### DIFF
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -40,9 +40,7 @@ In contrast to Cell Ranger, `cr-like-em` keeps multi-mapped reads and invokes an
 
 #### Combining counts from spliced cDNA and intronic regions
 
-For single-cell samples, we only included reads aligning to spliced and ambiguous cDNA transcripts in the counts matrix.
-For single-nuclei samples, all counts for spliced cDNA and intronic regions were summed for each gene to return the total counts summarized by gene in the counts matrix.
-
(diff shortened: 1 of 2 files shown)


jaclyn-taroni:
Now that we want to cut an initial release after this goes in, do we want to change the base branch to `main` @allyhawkins ?

allyhawkins:
> Now that we want to cut an initial release after this goes in, do we want to change the base branch to `main` @allyhawkins ?

Yes, I think that will be easier. I'll do that.

jashapiro (approved):
LGTM!

PR #180, opened 2023-10-31
Add note about download date at top of downloadable files section

jaclyn-taroni:
Closes #172. I'm unsure if I should be more specific since the illustrations will also be in this section (related: #171).

Issue #171, opened 2023-10-27
Update download illustrations to reflect appending download date to the top-most directory name


jaclyn-taroni:
#### Context

We're adding information about how to cite the portal in MLA and APA formats (#155). Folks are supposed to include the date on which they accessed a website in those formats. To make that easier for users, we will append dates to **the top-most directory names in downloads** (see _Dynamic filenames to include download date_ in https://github.com/AlexsLemonade/scpca-portal/issues/451#issuecomment-1779580195). Our current illustrations of the download structure don't reflect the addition of the date yet

#### Next steps

@dvenprasad please update [all illustrations currently in `development`](https://github.com/AlexsLemonade/scpca-docs/tree/727a183ede407f639ab84f3ad1d58e481d93fe78/docs/images) to include the date. For project downloads, `<project-accession-id>` would become `<project-accession-id>_<download-date>`; sample downloads will have a top-most directory that's `<sample-id>_<download-date>`.

We'll also track getting things into `main` and `development` on this ticket. I expect that the changes in https://github.com/AlexsLemonade/scpca-portal/issues/451 will be deployed before `AnnData` will be released, so this is how I think we should approach this:

 - [ ] All illustrations are updated and posted here. (@dvenprasad, pop into `#scpca-portal` on Slack and tag scientists so we can figure out who will pick up the rest of the ticket, please!)
 - [ ] `SingleCellExperiments` are updated
   - [ ] Create a feature branch off of `main` and add new 
   - [ ] Merge feature branch into `main`
   - [ ] Merge feature branch into `development`
 - [ ] `AnnData` images are updated in `development`

dvenprasad:
**Single-cell Project Download Folder**
![sc-project-download-folder](https://github.com/AlexsLemonade/scpca-docs/assets/15315514/a7d6baaf-ab9b-4e92-92a7-4a0a98528634)

**Single-cell Sample Download Folder**
![sc-sample-download-folder](https://github.com/AlexsLemonade/scpca-docs/assets/15315514/fc291781-b557-4eb7-aca3-b11301876b45)

**AnnnData Project Download Folder**
![anndata-project-download-folder](https://github.com/AlexsLemonade/scpca-docs/assets/15315514/989bcdf0-00fd-4aea-9927-1ab531c8f734)

**AnnData CITE seq Download Folder**
[...]

PR #181, opened 2023-11-01
Replace all downloadable file images currently in main with access date images

jaclyn-taroni:
⚠️ Stacked on #180 – see that PR for the sentence included at the beginning of the downloadable files section. 

Here, I'm addressing the part of #171 concerned with download illustrations already in `main`: project, sample, multiplex, and spatial. Ultimately, I've chosen to get this into `main` via the `documentation-improvements` branch, which contrasts what I wrote on the issue initially (written before our discussion on Friday). We will eventually merge `documentation-improvements` -> `main` and then `main` -> `development`, so I figured there was no harm in this approach.

The most important thing to check is that I didn't save an image under the wrong name!


docs/images/multiplexed-download-folder.png
allyhawkins, line 1: I think we might need to update this image. We have `processed.rds` files that are available for multiplexed data that is not reflected here. Looks like we missed updating that image.

allyhawkins (commented):
Everything looks fine, but I did notice that we are missing a file that should be in the multiplexed image. I think we should make sure to update that so that each sample folder also has `SCPCL000000_processed.rds`. Tagging @dvenprasad for help with that.

dvenprasad:
Here is the updated multiplexed download folder
multiplexed-download-folder

docs/images/multiplexed-download-folder.png
jaclyn-taroni: Updated in https://github.com/AlexsLemonade/scpca-docs/pull/181/commits/66562f16b5c26f172379ead8f8aa4c991d197efe

allyhawkins (approved):
Looks good!

PR #197, opened 2023-11-10
set release date 2023.11.10

davidsmejia:
Replaces the change log's first entry's title from Release Date placeholder to today's date: `2023.11.10`. This is the semver version of the date so it will be consistent with urls and the version on the side nav.


jashapiro (approved):
LGTM

PR #200, opened 2023-11-10
Merge change in Main to development

jashapiro:
There were a few conflicts when merging the changes that had occurred in https://github.com/AlexsLemonade/scpca-docs/tree/documentation-improvements back into the `development` branch, so out of an abundance of caution (and probably because it is required), I am filing that merge as a PR. There were not many conflicts, and only one that was resolved in the direction of the `main` branch, which was the change in `docs/sce_file_contents.md` where one says "Cell metrics" and the other "SingleCellExperiment Cell Metrics"


jashapiro:
I didn't see #199 before submitting this, but I think we want to do it this way anyway, rather than making changes to any files in `main`. But #199 may be a good reference for what those conflicts were!

docs/sce_file_contents.md
```diff
@@ -33,7 +34,9 @@ The `counts` assay can be accessed with the following R code:
 counts(sce) # counts matrix
 ```
 
-### SingleCellExperiment cell metrics
+Additionally, the `spliced` assay contains a counts matrix that includes reads from spliced cDNA only.
+
+### Cell metrics
```
allyhawkins, line 39:
```suggestion
### SingleCellExperiment Cell metrics
```
We actually want this to say `SingleCellExperiment`. That allows us to separate out `AnnData` and `SingleCellExperiment` instructions and is consistent with the rest of this document.

allyhawkins (approved):
This looks good, just one small change that needs to be fixed.

PR #203, opened 2023-11-15
Update FAQ for participant IDs

### LINKED ISSUE #190: Remove FAQ about missing participant IDs
When we add AnnData in the portal, we will also be adding participant IDs for all samples. Right now we have a [FAQ: Why do some samples have missing participant IDs? ](https://scpca.readthedocs.io/en/latest/faq.html#why-do-some-samples-have-missing-participant-ids), that should probably get updated. 

I think we might want to replace this with a FAQ on what is a participant ID.

allyhawkins:
Closes #190 

Here I updated the FAQ on why participant IDs might be missing to explain what a participant ID is. We now include a participant ID for every sample, so any language about the ID being optional has been removed.


jashapiro (approved):
LGTM... Do we want to say anything about where these come from?

allyhawkins:
> Yes, that is mostly what I meant; saying they were provided by submitters was the main point I was thinking of. I don't know that we need it, so up to you.

I added "provided by the submitter" in https://github.com/AlexsLemonade/scpca-docs/pull/203/commits/dfc3f6262f1f73c9a154433d8e7ed3a7d543f2da

PR #204, opened 2023-11-15
Brackets to parentheses

sjspielman:
As pointed out here, https://github.com/AlexsLemonade/scpca-docs/issues/165#issuecomment-1813235397, some link parentheticals are in brackets, not parentheses. 

> It seems pretty reasonable. I will pick a nit, though – some of the publication references in the cell annotation section use `[]` instead of `()` and I don't know why.

This PR fixes those spots! Did I miss any, or notice anything else while we're here?


jaclyn-taroni (approved):
Those were the two I spotted!

PR #207, opened 2023-11-17
Add new section for AnnData in Changelog

### LINKED ISSUE #189: Add changelog entry that includes information about added AnnData
The downloads from the portal will soon be available as either AnnData or SCE objects. The following should be noted in the changelog entry: 

- Data is now available as either SingleCellExperiment/RDS or AnnData/HDF5 files. This applies to all libraries except for multiplexed and spatial. 
- Ontology terms are now included in the metadata with each download 
- Cluster assignments are now available for libraries. I don't know if we should include this in the changelog because it means a new plot in the QC report, but do we want people to use these clusters? 🤷‍♀️ I'm just adding it to this list so we don't forget clusters are also being added. 
- Participant IDs have now been added for all samples. These used to be NA for most samples that only came from a single participant. 
- All sample metadata is now stored in the actual objects themselves (I'm 50/50 on if this is important enough for changelog. I think it's nice to tell users now that they can directly access metadata in the object since they couldn't do that before.) 

This entry should be added after the entry mentioned in #187.

allyhawkins:
Closes #189 

Here, I created a new entry in the Changelog that includes everything relevant to AnnData additions. I added a placeholder for the heading that will need to be replaced with the date of release when the time comes. 

- Are there any items that you think are not relevant to be in the Changelog entry? 
- Are there any items that should be there that are missing? (I didn't include mention of the clustering assignments, but I could)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -15,9 +15,9 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 ## PLACEHOLDER FOR RELEASE DATE
 
 * The data for all samples, other than multiplexed samples, can now be downloaded as either a `SingleCellExperiment(R)` or `AnnData(Python)` object contained in a RDS or HDF5 file, respectively.
-* The sample metadata has been updated to include ontology term ids for age, sex, organism, diagnosis, and tissue location.
-* All samples now have an assigned `participant_id`.
-* All data files now include both the gene expression data and metadata for each sample (e.g., age, sex, organism, diagnosis, and tissue location).
+* The sample metadata found in `single_cell_metadata.tsv` has been updated to include ontology term ids for age, sex, organism, ethnicity, diagnosis, and tissue location, when available.
+* All samples now have an assigned `participant_id`, which can be found in `single_cell_metadata.tsv`.
+* All data files now include both the gene expression data and metadata for each sample (e.g., age, sex, organism, ethnicity, diagnosis, and tissue location).
 For more information on the contents of the data files, see {ref}`the Single-cell gene expression file contents page<sce_file_contents:Single-cell gene expression file contents>`.
 * If provided by the original submitter, data files will also include cell type annotations.
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,15 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 <!-- PUT THE NEW CHANGELOG ENTRY RIGHT BELOW THIS -->
 <!-------------------------------------------------->
 
+## PLACEHOLDER FOR RELEASE DATE
+
+* The data for all samples, other than multiplexed samples, can now be downloaded as either a `SingleCellExperiment(R)` or `AnnData(Python)` object contained in a RDS or HDF5 file, respectively.
+* The sample metadata has been updated to include ontology term ids for age, sex, organism, diagnosis, and tissue location.
+* All samples now have an assigned `participant_id`.
+* All data files now include both the gene expression data and metadata for each sample (e.g., age, sex, organism, diagnosis, and tissue location).
+For more information on the contents of the data files, see {ref}`the Single-cell gene expression file contents page<sce_file_contents:Single-cell gene expression file contents>`.
+* If provided by the original submitter, data files will also include cell type annotations.
+
 ## 2023.11.10
 
 * The README included in your download now contains the following:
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
...
 <!-------------------------------------------------->
 
+## PLACEHOLDER FOR RELEASE DATE
+
+* The data for all samples, other than multiplexed samples, can now be downloaded as either a `SingleCellExperiment(R)` or `AnnData(Python)` object contained in a RDS or HDF5 file, respectively.
+* The sample metadata found in `single_cell_metadata.tsv` has been updated to include ontology term ids for age, sex, organism, ethnicity, diagnosis, and tissue location, when available.
+* All samples now have an assigned `participant_id`, which can be found in `single_cell_metadata.tsv`.
+* All data files now include both the gene expression data and metadata for each sample (e.g., age, sex, organism, ethnicity, diagnosis, and tissue location).
+For more information on the contents of the data files, see {ref}`the Single-cell gene expression file contents page<sce_file_contents:Single-cell gene expression file contents>`.
+* If provided by the original submitter, data files will also include cell type annotations.
```
jaclyn-taroni, line 22:
```suggestion
* Data files will include cell type annotations provided by submitters when applicable.
```

docs/CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 <!-- PUT THE NEW CHANGELOG ENTRY RIGHT BELOW THIS -->
 <!-------------------------------------------------->
 
+## PLACEHOLDER FOR RELEASE DATE
+
+* The data for all samples, other than multiplexed samples, can now be downloaded as either a `SingleCellExperiment(R)` or `AnnData(Python)` object contained in a RDS or HDF5 file, respectively.
```
jaclyn-taroni, line 17: Is there a section we should link to after the first section as well?

```suggestion
* Downloads for most projects are now available in [`AnnData`](https://anndata.readthedocs.io/en/latest/index.html) format as HDF5 files.
Multiplexed samples are not yet supported.
```

I don't think we need to state that the `SingleCellExperiment` downloads are available.

docs/CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
...
+* The sample metadata found in `single_cell_metadata.tsv` has been updated to include ontology term ids for age, sex, organism, ethnicity, diagnosis, and tissue location, when available.
+* All samples now have an assigned `participant_id`, which can be found in `single_cell_metadata.tsv`.
```
jaclyn-taroni, line 19: I think it'd be helpful to give more information about what we _used to do_ to provide more context for the potential impact:
```suggestion
* All samples now have an assigned `participant_id`, which can be found in `single_cell_metadata.tsv`.
Previously, `participant_id` were only assigned when multiple samples mapped to the same participant for most projects.
```

docs/CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 <!-- PUT THE NEW CHANGELOG ENTRY RIGHT BELOW THIS -->
 <!-------------------------------------------------->
 
+## PLACEHOLDER FOR RELEASE DATE
+
+* The data for all samples, other than multiplexed samples, can now be downloaded as either a `SingleCellExperiment(R)` or `AnnData(Python)` object contained in a RDS or HDF5 file, respectively.
+* The sample metadata found in `single_cell_metadata.tsv` has been updated to include ontology term ids for age, sex, organism, ethnicity, diagnosis, and tissue location, when available.
```
jaclyn-taroni, line 18: Is there a section for each download type we could link to here?

jaclyn-taroni (commented):
Thanks for getting this going! My general comment is that a few of these points would benefit from some links to sections of the docs when possible.

docs/CHANGELOG.md
allyhawkins: This is referring to the metadata table that's included in the download, not the metadata that's in the actual object. I added a link to that section of the docs.

allyhawkins:
Thanks for looking at this @jaclyn-taroni. I incorporated your wording suggestions and then also added a link to the metadata section of the downloads for the bullet about added metadata included in the download. This should be ready for another look.

jaclyn-taroni (approved):

PR #233, opened 2024-01-12
Change one more name

sjspielman:
I changed `_qc_report.html -> _qc.html`.
FYI no parallel change needed in `scpca-nf` - https://github.com/AlexsLemonade/scpca-nf/blob/f08ce452f95f88e169217c114aef1c1321a3688c/modules/qc-report.nf#L17

### DIFF
--- a/docs/download_files.md
+++ b/docs/download_files.md
@@ -77,7 +77,7 @@ See also {ref}`Using the provided RDS files in R <faq:how do i use the provided 
 
 ## QC report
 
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
This is fine, but we should add this to the anndata release, so the next release it gets incorporated.

sjspielman:
> This is fine, but we should add this to the anndata release, so the next release it gets incorporated.

Ahh. I'll cherry-pick it into that branch (and cell type)!

PR #208, opened 2023-11-20
Update docs in `main` for AnnData release

allyhawkins:
## Purpose

Here I am adding all the documentation that is needed for the addition of AnnData downloads. 
This has all the updates in the docs that were made that are specific to AnnData and includes a new changelog entry for these additions. 
I have not yet added a date to the changelog, since we aren't fully ready to release the AnnData changes. So before this PR can go in, we will need to update the date. I'm filing this as a draft for now, and once the portal is ready we can update the changelog and get this in. 

### Issue addressed

#202 

## Checklist for publishing to `main`

- [x] This branch contains all commits/content ready to be released.
- [x] The most recent CHANGELOG entry includes a date.
If this release required a CHANGELOG entry, it should be today's date.

PR #236, opened 2024-01-16
Fix typo: filtered -> processed

sjspielman:
While working on merged docs, I noticed this typo where filtered should have said processed.

I fixed this typo separately since it needs to make it into #208 for the anndata release; I can cherry-pick it in there (and cell type release PR) after this is goes in.


allyhawkins (approved):
🐱

PR #242, opened 2024-01-19
No cell type report download for cell lines

sjspielman:
**Stacked on cell type release PR**

The current chat in Slack reminded me that we need to tell users that they won't receive a report with their download if the library is from a cell line! My wording might be simplified as well here, lmk what you think.


allyhawkins (approved):
🐱 ❄️

PR #243, opened 2024-01-19
Add FAQ for libraries that get removed

### LINKED ISSUE #143: Add FAQ about samples/libraries getting removed
We generally take an approach where we do not strictly pass or fail samples for release based on QC stats. Instead, we release the samples with the QC report. However, we also respect investigators' requests to remove poor-quality samples. In the past, that removal has happened before release. There are some upcoming cases where samples might get removed per request after release. It might be helpful to add an FAQ explaining why samples "disappear," and frame it in the context of our overall approach to QC.

allyhawkins:
## Purpose

Here I'm adding a FAQ to explain why some libraries were previously there but might be missing. I tried to focus on the quality control and mentioned that we do not do any screening. 

I'm also filing this to `main` based on our discussion in Slack. We don't need to release this immediately, but it will now be included in the next release. If we prefer to add it to `development` and then to the AnnData release branch, I can move it there. 

### Issue addressed

Closes #143 

## Checklist for publishing to `main`

- [x] This branch contains all commits/content ready to be released.
- [ ] The most recent CHANGELOG entry includes a date.
If this release required a CHANGELOG entry, it should be today's date.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -214,10 +214,8 @@ Alternatively, you might be interested in this [reference from the authors of `s
 Although we try to maintain backward compatibility, new features added to the ScPCA Portal may result in downloads that are no longer compatible with code written with older downloads from the ScPCA Portal in mind.
 Please see our {ref}`CHANGELOG <CHANGELOG:CHANGELOG>` for a summary of changes that impact downloads from the Portal.
 
-## I previously downloaded a sample that is no longer on the Portal, why can't I find it?
-
-We process and release all data provided to us on the Portal.
-We do not conduct any quality screening before releasing the data.
-However, to assess library quality, we provide a QC report with each download.
-There have been instances where submitters have conducted their own analysis and discovered low-quality libraries after they have been released on the Portal.
-If they request to remove those libraries and associated samples, we respect their wishes and that sample will no longer be available.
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -213,3 +213,11 @@ Alternatively, you might be interested in this [reference from the authors of `s
 
 Although we try to maintain backward compatibility, new features added to the ScPCA Portal may result in downloads that are no longer compatible with code written with older downloads from the ScPCA Portal in mind.
 Please see our {ref}`CHANGELOG <CHANGELOG:CHANGELOG>` for a summary of changes that impact downloads from the Portal.
+
+## I previously downloaded a sample that is no longer on the Portal, why can't I find it?
+
+We process and release all data provided to us on the Portal.
+We do not conduct any quality screening before releasing the data.
+However, to assess library quality, we provide a QC report with each download.
```
jaclyn-taroni, line 221:
```suggestion
Instead, we provide a QC report with each download so that users may assess library quality themselves.
```

jaclyn-taroni (commented):
The overall content looks good, but I think we can be more straight to the point by moving up _why_ the sample is no longer available.

allyhawkins:
I updated this to reflect your changes @jaclyn-taroni so should be ready for another look.

jaclyn-taroni (approved):
👍

PR #250, opened 2024-01-23
Change some language from libraries to samples

### LINKED ISSUE #248: Update merged processing docs to focus on samples, not just libraries
I wonder if we want to update the processing information section for merged objects to also mention combining data from all libraries and all samples. In the rest of the processing information docs (and the download section), we talk in the context of samples, so I think we want to make sure people understand that the data from all samples in a project is what is used to create that merged object.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/scpca-docs/pull/244#pullrequestreview-1836835655_

sjspielman:
Closes #248 

This PR does what it says and changes some "library" -> "sample" in the processing docs for merged objects. I really only changed spots in the parts about downloading files, since it seemed to me that library _might_ be more appropriate in the meat of the processing description. For review, please give the whole section a quick skim (including parts not shown in the diff) and lmk if I should update other spots too, thanks!

### DIFF
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -181,8 +181,8 @@ We chose to enable the [`--seqBias`](https://salmon.readthedocs.io/en/latest/sal
 
 ## Merged objects
 
-In addition to providing separate objects for each library, we also offer an option to download all files for a project as a single merged object.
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
I read the whole section with these changes and I think this is sufficient!

PR #223, opened 2024-01-03
Update docs in `main` for cell type release

sjspielman:
## Purpose

This PR contains all changes necessary for the cell type docs release. 
Given the current in-progress `AnnData` release in #208, this PR was created by branching off of [`allyhawkins/anndata-release`](https://github.com/AlexsLemonade/scpca-docs/tree/allyhawkins/anndata-release) and merging in `development`, as no cherry-picking was necessary.

This PR is being filed as a draft to `main`, since that is the ultimate target branch. In order to mark it ready for formal review and de-draft, these changes will first need to be made:

- Merge in `main` after #208 is merged; this will provide the `AnnData` release `CHANGELOG` date and keep us in sync with `main`.
- Update the `CHANGELOG` with this cell type release date (note that I'm leaving the associated issue #209 open until this is finalized)

### Issue addressed

- `CHANGELOG` issue: #209
- Release issue: #210

## Checklist for publishing to `main`

- [x] This branch contains all commits/content ready to be released.
- [x] The most recent CHANGELOG entry includes a date.
If this release required a CHANGELOG entry, it should be today's date.

PR #253, opened 2024-01-24
Add sample_type metadata

### LINKED ISSUE #252: Update sce & merged contents with sample type
From https://github.com/AlexsLemonade/scpca-nf/issues/645#issuecomment-1906731450

> > Given this, I'm opening up a quick chat for how folks feel about also aiming to get this information into the SCE in the first place; for example, something like metadata(sce)$sample_type <- <"cell line"/"patient-derived xenograft"/"patient tissue">, in the generate_unfiltered_sce area of the code. @allyhawkins @jashapiro
> 
> Since we are already reading in the sample metadata in this script, then I would agree that this is a good place to add that information.

We're going to be added a new field into the SCE metadata. This issue tracks documenting it here. 
Once this is actually implemented in `scpca-nf`, I'll come back and comment here the specific docs changes we'll need.

sjspielman:
Closes #252 

Here, I'm adding `sample_type` to the file content descriptions `metadata` sections. I realize also that I have a VS Code plugin for working with markdown tables that changed up the spacing a bit, so as a consequence the whole tables are showing a diff!

But I really only changed (added) 1 line! I added it right after `sample_metadata`, which corresponds to where it gets added in the workflow.

```
| `sample_type`   | A string indicating the type of sample, with one of the following values: `"patient-derived xenograft"`, `"cell line"`, or `"patient tissue"`. If the library is multiplexed, this will be a named vector giving the sample type for each sample ID in the object   |
```

After this goes into `development`, I'm going to cherry-pick into #223


docs/merged_objects.md
```diff
@@ -132,44 +132,45 @@ metadata(merged_sce)$library_metadata$SCPCL000000
...
+| `sample_metadata`              | Data frame containing metadata for each sample included in the library (see the [`Sample metadata` section below](#singlecellexperiment-sample-metadata))                                                                                                                                                                                                                          |
+| `sample_type`                  | A string indicating the type of sample, with one of the following values: `"patient-derived xenograft"`, `"cell line"`, or `"patient tissue"`. If the library is multiplexed, this will be a named vector giving the sample type for each sample ID in the object                                                                                                                  |
```
allyhawkins, line 152:
```suggestion
| `sample_type`                  | A string indicating the type of sample, with one of the following values: `"patient-derived xenograft"`, `"cell line"`, or `"patient tissue"`. If the library is multiplexed, this will be a named vector giving the sample type for each sample ID in the library                                                                                                                  |
```
I think we want to say library here, because all of this information will be in a nested list by library ID in the merged object.

docs/sce_file_contents.md
```diff
@@ -118,47 +118,48 @@ Metadata associated with {ref}`data processing <processing_information:Processin
...
+| `transcript_type`              | Transcripts included in gene counts: `spliced` for single-cell samples and `unspliced` for single-nuclei                                                                                                                                                                                                                                                                           |
+| `sample_metadata`              | Data frame containing metadata for each sample included in the library (see the [`Sample metadata` section below](#singlecellexperiment-sample-metadata))                                                                                                                                                                                                                          |
+| `sample_type`                  | A string indicating the type of sample, with one of the following values: `"patient-derived xenograft"`, `"cell line"`, or `"patient tissue"`. If the library is multiplexed, this will be a named vector giving the sample type for each sample ID in the object  |
```
allyhawkins, line 138:
```suggestion
| `sample_type`                  | A string indicating the type of sample, with one of the following values: `"patient-derived xenograft"`, `"cell line"`, or `"patient tissue"`. If the library is multiplexed, this will be a named vector giving the sample type for each sample ID in the library  |
```
Same thing, since we're talking about multiplexed libraries, let's keep the word library and not object.

allyhawkins (approved):
This looks fine and I'm going to approve as long as you take my suggestion to use `library` instead of `object`.

PR #255, opened 2024-01-25
Add section to FAQ about projects without merged objects

### LINKED ISSUE #221: FAQ: Why is a merged object unavailable for this project?
You can't get a merged object if you're downloading a spatial project download. We must explain why a merged object is unavailable in these cases and any others. The other example I'm aware of is AnnData project downloads when a project _exclusively_ comprises multiplexed samples, but I don't believe we currently have anything that fits that criterion.

sjspielman:
Closes #221

In this PR and very well-named branch, I added a section to the FAQ about circumstances where there is no merged object. I think this could be two sections instead of one, but since each would be so small I started off by collapsing into one section. 

Please let me know what you think of the section title, whether it should be 1 or 2 sections, and whether I've hit all the right points or should expand/reduce anywhere!


allyhawkins (approved):
Honestly I think this looks good and I like how you've phrased things!

Issue #241, opened 2024-01-19
 Add AnnData sections to merged objects docs page


sjspielman:
_Broken out of #219_

The scope of this issue is wrapping up all `AnnData` object contents in the `merged_objects.md` docs page.

sjspielman:
Closed with #260 and #262 🚀

PR #260, opened 2024-01-31
Add most of AnnData merged docs

sjspielman:
Towards #241 
This is the first of two PRs (second coming after https://github.com/AlexsLemonade/scpca-nf/issues/681 is wrapped up) to fill in the rest of the `AnnData` docs for merged objects. 

Note that I have one `!!TODO!` little bit in there since whether we end up with `suspension_type` (vs just keeping it as `seq_unit`) depends on that https://github.com/AlexsLemonade/scpca-nf/issues/681 as well. I imagine we'll be keeping it as `seq_unit` since we don't need CZI compliance for merged objects?

There are also a few corrections I made along the way to `SCE` merged objects that I noticed:
- We were missing the `additional_modalities` column in `colData`
- The `metadata(merged_sce)$library_metadata$SCPCL000000` list was missing a few fields, and it also incorrectly listed `sample_metadata` which was removed. 
- Fixed a couple typos, also in `sce_object_contents.md` where the same typo was in both files.
- Also, in `sce_object_contents.md`, `suspension_type` was missing from the `AnnData` `obs` slot, and it indeed appears to be in those objects (checked on in `scpca-prod`).


docs/merged_objects.md
```diff
@@ -134,7 +132,9 @@ Each such list will contain the following fields:
...
+| `library_id`                   | Library ID in the form `SCPCL000000`                                                                                                                                                                                                                                                                                                                                               || `salmon_version`               | Version of `salmon` used for initial mapping
+| `project_id`                   | Project ID in the form `SCPCP000000`                                                                                                                                                                                                                                                                                                                                               || `salmon_version`               | Version of `salmon` used for initial mapping
```
allyhawkins, line 137:
```suggestion
| `project_id`                   | Project ID in the form `SCPCP000000`                                                                                                                                                                                                                                                                                                                                               |
```
We shouldn't have this twice?

allyhawkins: Is this `project_id` or `scpca_project_id`?

docs/merged_objects.md
```diff
@@ -368,4 +365,63 @@ merged_adata_object.obs_names # matrix column names
...
+| ------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `sample_id`                                | Sample ID in the form `SCPCS000000`                                                                                                                                                                                                       |
+| `scpca_project_id`                         | Project ID in the form `SCPCP000000`                                                                                                                                                                                                      |
+| `library_id`                               | Library ID in the form `SCPCL000000`                                                                                                                                                                                                      |
```
allyhawkins, line 387: Don't we have these in the SCE objects? So they don't need to be included here.

docs/merged_objects.md
```diff
@@ -368,4 +365,63 @@ merged_adata_object.obs_names # matrix column names
...
+| `tissue_ontology_term_id`                  | [`Uberon`](http://obofoundry.org/ontology/uberon.html) term indicating tissue of origin. If unavailable, `NA` is used                                                                                                                     |
+| `assay_ontology_term_id`                   | A string indicating the [Experimental Factor Ontology](https://www.ebi.ac.uk/ols/ontologies/efo) term id associated with the technology and version used for the single-cell library, such as 10Xv2, 10Xv3, or 10Xv3.1                    |
+| `seq_unit`                          | `cell` for single-cell samples or `nucleus` for single-nuclei samples                                                                                                                                                                     |
+| `is_primary_data`                          | Set to `FALSE` for all libraries to reflect that all libraries were obtained from external investigators. Required by `CELLxGENE`                                                                                                         |
```
allyhawkins, line 408: I think we won't have this?

docs/merged_objects.md
```diff
@@ -368,4 +365,63 @@ merged_adata_object.obs_names # matrix column names
...
+
+All of the per-gene data columns included in the `rowData` of the `SingleCellExperiment` objects are present in the `.var` slot of the `AnnData` object.
+Note that the `SingleCellExperiment` columns named `SCPCL000000-mean` and `SCPCL000000-detected` are instead named `SCPCL000000.mean` and `SCPCL000000.detected`, respectively, in the merged `AnnData` object.
+To see a full description of the included columns, see the [section on gene metrics in `Components of a SingleCellExperiment object`](#singlecellexperiment-gene-information-and-metrics).
+
+The `AnnData` object also includes the following additional gene-level metadata column:
 
+| Column name   | Contents                                                         |
+| ------------- | ---------------------------------------------------------------- |
+| `is_feature_filtered` | Boolean indicating if the gene or feature is filtered out in the normalized matrix but is present in the raw matrix     |
```
allyhawkins, line 427: Just noting that I think we won't have this either.

allyhawkins (commented):
Just a few comments to start, but I want to wait to fully review this until https://github.com/AlexsLemonade/scpca-nf/issues/681 is addressed.

docs/merged_objects.md
sjspielman: It's definitely `project_id`: 

```
> metadata(sce)$library_metadata$SCPCL000001 |> names()
 [1] "library_id"              "sample_id"              
 [3] "project_id"              "salmon_version"         
 [5] "reference_index"         "total_reads"            
 [7] "mapped_reads"            "mapping_tool"           
 [9] "alevinfry_version"       "af_permit_type"         
[11] "af_resolution"           "usa_mode"               
[13] "af_num_cells"            "tech_version"           
[15] "assay_ontology_term_id"  "seq_unit"               
[17] "transcript_type"         "include_unspliced"      
[19] "filtering_method"        "miQC_model"             
[21] "prob_compromised_cutoff" "scpca_filter_method"    
[23] "min_gene_cutoff"         "normalization"          
[25] "highly_variable_genes" 
```

But I poked around a little at some files in `scpca-prod` and I see:

1) `SCPCL000001_processed.rds`: `project_id` is in `metadata(sce)`
2) `SCPCL000001_processed_rna.hdf5`: `scpca_project_id` is in `obj.obs`

Those files were both produced with `v0.7.1`. I'm not totally sure where this difference came from...?

allyhawkins: It's because we add `project_id` to the metadata when we actually create the object using `scpcaTools::read_alevin()` and then when we add the sample metadata to the `colData` prior to AnnData conversion, the columns are coming from `scpca-sample-metadata.tsv`. I just wanted to make sure that discrepancy was correct.

sjspielman: > then when we add the sample metadata to the colData prior to AnnData conversion, the columns are coming from scpca-sample-metadata.tsv

Great, this is what I wasn't seeing! The discrepancy is correct.

sjspielman:
https://github.com/AlexsLemonade/scpca-nf/issues/681 is closed, so re-requesting review now.

docs/merged_objects.md
```diff
@@ -134,7 +126,9 @@ Each such list will contain the following fields:
...
+| `library_id`                   | Library ID in the form `SCPCL000000`                                                                                                                                                                                                                                                                                                                                               || `salmon_version`               | Version of `salmon` used for initial mapping
+| `project_id`                   | Project ID in the form `SCPCP000000`                                                                                                                                                                                                                                                                                                                                               |
```
allyhawkins, line 131:
```suggestion
| `library_id`                   | Library ID in the form `SCPCL000000`                                                                                                                                                                                                                                                                                                                                               |
| `project_id`                   | Project ID in the form `SCPCP000000`                                                                                                                                                                                                                                                                                                                                               |
| `salmon_version`               | Version of `salmon` used for initial mapping
|
```
I feel like this order should be switched, even if it's not this order in the object this makes more sense to me and order shouldn't matter in the actual object.

allyhawkins (approved):
Looks good, just one minor comment.

PR #271, opened 2024-02-27
No tumor in reference data

### LINKED ISSUE #270: Add sentence in processing information about cell type references being normal
> I think adding a sentence to the section on cell typing in the processing docs would be sufficient, or if we want to make sure users see it we could include it in the report itself.

both? Just a quick, "Please be aware that all cell type references include only normal (not tumor) tissue." would suffice, I think.

_Originally posted by @sjspielman in https://github.com/AlexsLemonade/ScPCA-admin/issues/716#issuecomment-1966685634_

sjspielman:
Closes #270 

I've added a quick caveat sentence to the docs here.
I'll cherry-pick this into the cell type release once it goes into `development`.

### DIFF
--- a/docs/processing_information.md
+++ b/docs/processing_information.md
@@ -84,7 +84,7 @@ We combine these marker genes with all "immune cell" `PanglaoDB` marker genes to
 During annotation, we additionally include an `"other"` cell type that does not express any of these marker genes.
 As a consequence, cells which `CellAssign` cannot confidently annotate from the full marker gene list are labeled as `"other"`.
 
-Please be aware that all cell type annotation reference datasets include only normal (not tumor) tissue.
+Please be aware that all cell type annotation reference datasets are derived from normal (not tumor) tissue.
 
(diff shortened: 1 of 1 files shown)


docs/processing_information.md
```diff
@@ -84,6 +84,8 @@ We combine these marker genes with all "immune cell" `PanglaoDB` marker genes to
 During annotation, we additionally include an `"other"` cell type that does not express any of these marker genes.
 As a consequence, cells which `CellAssign` cannot confidently annotate from the full marker gene list are labeled as `"other"`.
 
+Please be aware that all cell type annotation reference datasets include only normal (not tumor) tissue.
```
allyhawkins, line 87:
```suggestion
Please be aware that all cell type annotation reference datasets are derived from normal (not tumor) tissue.
```
How about this?

sjspielman: good with me!

allyhawkins (approved):
🐱

PR #272, opened 2024-02-27
Sync up development

sjspielman:
While cherry-picking #271 into the cell type release branch, I noticed a couple changes had apparently happened in the cell type release branch only which were not properly reflected in `development`. This PR merges the cell type branch into `development`, so branches should now be more in sync!


allyhawkins (approved):
🐱 🐱

PR #273, opened 2024-03-05
remove miQC_model from processed & merged

jashapiro:
In conjunction with https://github.com/AlexsLemonade/scpca-nf/pull/712, here I am updating docs to reflect that the `processed` and merged objects will no longer have `miQC_model` objects in the metadata.


sjspielman (approved):
👍

PR #279, opened 2024-03-12
Update cell type release docs

### LINKED ISSUE #278: Update cell type release for miQC & file compression changes
We have removed miQC object from processed objects and swapped to a different compression for rds files.
We want this to be part of the cell type release (#223)

- Add to CHANGELOG
  - no miQC objects in processed
  - you may notice slower read times for smaller files now since we're using bz2
- cherry-pick https://github.com/AlexsLemonade/scpca-docs/pull/273/commits/4952ef733bf5f82cb526e40c14f1412f206ea5a8 into celltype release branch

sjspielman:
Closes #278 

This PR adds a few items to the celltype release.

Sort of amusingly, when I cherry-picked in the `miQC_model` change, there was no real diff there! What gives, I wondered? Turns out this is based on an old typo where we were missing `processed` in that description! So, kinda worked out more smoothly.

So, the main thing to review here is the CHANGELOG.


allyhawkins (approved):
LGTM

PR #283, opened 2024-04-04
Mention that no merged objects are provided for larger projects

### LINKED ISSUE #282: Update FAQ to reflect projects with > 50 samples do not have merged objects
For now, we are not planning on releasing merged objects for larger projects. Currently, any projects with > 50 samples will not have a merged object. This scenario should be listed as a reason why a project is missing a merged object in the FAQ.

allyhawkins:
Closes #282 

Here I updated the docs to account for projects with missing merged objects. 

- I noticed that we still describe the contents of multiplexed merged objects even though we don't provide them. I removed that section and replaced with a note that we don't merge any projects with multiplexed libraries. 
- I updated the FAQ with another bullet for large projects. I used a cut-off of 50 since both projects with > 50 samples failed and projects with 45 samples were created without issues. I just mentioned that we don't include them because they are too large and hard to work with.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -192,6 +192,10 @@ There are two types of projects for which merged objects are not available:
     - Although the ScPCA pipeline {ref}`reports demultiplexing results<processing_information:HTO demultiplexing>`, it does not actually perform demultiplexing.
     As there is no guarantee that a unique HTO was used for each sample in a given project, it would not necessarily be possible to determine which HTO corresponds to which sample in a merged object.
 
+- Projects containing more than 50 samples
+    - The more samples that are included in a merged object, the larger the object, and the more difficult it will be to work with that object in R or Python.
+    Because of this, we do not provide merged objects for projects with more than 50 samples as the size of the merged object is too large.
+
 ## Why doesn't my existing code work on a new download from the Portal?
 
 Although we try to maintain backward compatibility, new features added to the ScPCA Portal may result in downloads that are no longer compatible with code written with older downloads from the ScPCA Portal in mind.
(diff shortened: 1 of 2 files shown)


docs/merged_objects.md
```diff
@@ -311,28 +311,8 @@ metadata(altExp(merged_sce, "adt")) # adt metadata
...
-| --------------------------- | ---------------------------------------------------------------- |
-| `hashedDrops_sampleid`      | Most likely sample as called by `DropletUtils::hashedDrops`      |
-| `HTODemux_sampleid`         | Most likely sample as called by `Seurat::HTODemux`               |
-| `vireo_sampleid`            | Most likely sample as called by `vireo` (genetic demultiplexing) |
-
-
-Unlike in {ref}`individual SingleCellExperiment objects<sce_file_contents:additional SingleCellExperiment components for multiplexed libraries>`, hashtag oligo (HTO) quantification will not be included in the merged `SingleCellExperiment` as an alternative experiment, as described in the ref`{frequently asked questions:faq:which projects can I download as a merged objects?}>`.
-
+Merged objects are not available for any projects that contain multiplexed libraries.
+As there is no guarantee that a unique HTO was used for each sample in a given project, it would not necessarily be possible to determine which HTO corresponds to which sample in a merged object.
```
sjspielman, line 315:
```suggestion
This is because there is no guarantee that a unique HTO was used for each sample in a given project, so it would not necessarily be possible to determine which HTO corresponds to which sample in a merged object.
```

docs/faq.md
```diff
@@ -192,6 +192,10 @@ There are two types of projects for which merged objects are not available:
     - Although the ScPCA pipeline {ref}`reports demultiplexing results<processing_information:HTO demultiplexing>`, it does not actually perform demultiplexing.
     As there is no guarantee that a unique HTO was used for each sample in a given project, it would not necessarily be possible to determine which HTO corresponds to which sample in a merged object.
 
+- Projects containing more than 50 samples
```
sjspielman, line 195: On line 185, need to change `two` -- `three` (can't comment directly)

sjspielman (approved):
LGTM, just two small changes!

PR #286, opened 2024-04-07
Multiplexed download folder image does not have a celltype report.

davidsmejia:
## Purpose

This PR is based off of @sjspielman #223

I assume that we should have multiplexed celltype reports in the input bucket. But we don't at the moment for samples that are not demuxed. Seeing as everyone is out for AACR. I'd like to deploy to production and start processing but it seems as though the docs are out of sync with that will be on the portal.

Since multiplexed celltype reports are not currently available in the input bucket. I have filed this PR to temporarily show the multiplexed download structure that will reflect what the users will download.

Note: Single cell samples in the multiplexed project do contain a celltype report and are accurately covered in the docs.

Also updates CHANGELOG to Sunday April 7th

I added the old image with a new name because I don't know where those images come from and it seemed reasonable. 

### Issue addressed

No Issue filed yet for missing multiplexed 

## Checklist for publishing to `main`

- [x] This branch contains all commits/content ready to be released.
- [X] The most recent CHANGELOG entry includes a date.
If this release required a CHANGELOG entry, it should be today's date.


davidsmejia:
Note: the corresponding readme that is included with multiplexed samples has been updated for now to not show cell type reports in those downloads.

allyhawkins:
@davidsmejia I looked into this since this is not correct and it looks like there might be an issue with the hand off script not properly copying over data for multiplexed samples. So that data is not up to date with what we have for processing. I just updated the hand off bucket so it now has the most up to date objects for multiplexed including the cell type report. The staging bucket should also be updated.

PR #290, opened 2024-04-12
Change threshold for merged objects from 50 to 100

allyhawkins:
Here I'm changing the max number of samples to create a merged object from 50 to 100.


jashapiro (approved):
📈

Issue #294, opened 2024-04-19
Add CHANGELOG for transition to `.h5ad` extensions


jaclyn-taroni:
Change includes `.hdf5` -> `.h5ad` in AnnData downloads.

jashapiro:
This is already present in `development`: https://github.com/AlexsLemonade/scpca-docs/blob/development/docs/CHANGELOG.md

jaclyn-taroni:
I will take this opportunity to state that this transition didn't involve enough issues for my taste.

PR #295, opened 2024-04-19
Update CHANGELOG.md placehodler header

jashapiro:
I realized that while the bullet point text required for #294 was present, the placeholder header was not updated properly, so here I am doing that.


jaclyn-taroni (approved):
I was going to file this but you beat me!

PR #298, opened 2024-04-23
FAQ for why AnnData isn't available

### LINKED ISSUE #297: Add FAQ: Why aren't some samples available as AnnData
This should cover which samples aren't available as AnnData (spatial and multiplexed) and why.

allyhawkins:
Closes #297 

This PR adds a FAQ for why some samples are unavailable to download as AnnData objects. I used the same overall structure that we use to explain why some merged objects aren't available. I was struggling with the explanation of why we don't include AnnData for multiplexed libraries, so any guidance there would be greatly appreciated!

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -64,10 +64,9 @@ There are two types of samples where `AnnData` objects are not available:
     Therefore, we only provide the output from running Space Ranger and do not store data in either R or Python objects.
 
 - Samples that are part of multiplexed libraries
-    - Although the ScPCA pipeline {ref}`reports demultiplexing results<processing_information:HTO demultiplexing>`, it does not actually perform demultiplexing.
-    `AnnData` objects do not support storage of both gene expression data and additional modalities, like quantified HTO expression data, within the same object.
-    Without the associated HTO data, the sample of origin cannot be identified for each cell, causing ambiguity in sample origin assignments.
-    Therefore, we do not provide any multiplexed libraries as `AnnData` objects.
+    - Although the ScPCA pipeline {ref}`reports demultiplexing results<processing_information:HTO demultiplexing>`, it does not definitively separate samples due to the potential for disagreement among methods.
+    - Resolving such disagreements requires examination of the HTO data, which can not be stored in the same `AnnData` object. 
+    Therefore, we do not currently provide any multiplexed libraries as `AnnData` objects.
 
 ## What is the difference between samples and libraries?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -65,9 +65,9 @@ There are two types of samples where `AnnData` objects are not available:
 
 - Samples that are part of multiplexed libraries
     - Although the ScPCA pipeline {ref}`reports demultiplexing results<processing_information:HTO demultiplexing>`, it does not definitively separate samples due to the potential for disagreement among methods.
-    - Resolving such disagreements requires examination of the HTO data, which can not be stored in the same `AnnData` object. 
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -54,6 +54,21 @@ scpca_sample = anndata.readh5ad(file = "SCPCL000000_processed_rna.h5ad")
...
+
+- Spatial transcriptomics data from a given sample
+    - As described in {ref}`the spatial transcriptomics processing section<processing_information:spatial transcriptomics>`, no post-processing is performed on these libraries after running Space Ranger.
+    Therefore, we only provide the output from running Space Ranger and do not store data in either R or Python objects.
+
+- Samples that are part of multiplexed libraries
+    - Although the ScPCA pipeline {ref}`reports demultiplexing results<processing_information:HTO demultiplexing>`, it does not actually perform demultiplexing.
+    `AnnData` objects do not support storage of both gene expression data and additional modalities, like quantified HTO expression data, within the same object.
+    Without the associated HTO data, the sample of origin cannot be identified for each cell, causing ambiguity in sample origin assignments.
+    Therefore, we do not provide any multiplexed libraries as `AnnData` objects.
```
jashapiro, line 70: To be honest, there probably isn't a good reason not to have AnnData for multiplexed libraries, though it would require us to figure out how to make tables of results. I think the main reason may be that anything we would produce would not be compatible with CellXGene? At least that is my recollection of the initial logic. Not doing the merged samples is a bit easier to explain, because of the challenge of mapping reused HTO tags, but that doesn't apply here. 

Here is my attempt:

```suggestion
    - Although the ScPCA pipeline {ref}`reports demultiplexing results<processing_information:HTO demultiplexing>`, it does not definitively separate samples due to the potential for disagreement among methods.
    - Resolving such disagreements requires examination of the HTO data, which can not be stored in the same `AnnData` object. 
    Therefore, we do not currently provide any multiplexed libraries as `AnnData` objects.
```

jashapiro: > Yes, this is exactly why we don't have them which is why I was struggling. 

Do we maybe want to say this explicitly then?

docs/faq.md
```diff
@@ -54,6 +54,20 @@ scpca_sample = anndata.readh5ad(file = "SCPCL000000_processed_rna.h5ad")
...
+There are two types of samples where `AnnData` objects are not available:
+
+- Spatial transcriptomics data from a given sample
+    - As described in {ref}`the spatial transcriptomics processing section<processing_information:spatial transcriptomics>`, no post-processing is performed on these libraries after running Space Ranger.
+    Therefore, we only provide the output from running Space Ranger and do not store data in either R or Python objects.
+
+- Samples that are part of multiplexed libraries
+    - Although the ScPCA pipeline {ref}`reports demultiplexing results<processing_information:HTO demultiplexing>`, it does not definitively separate samples due to the potential for disagreement among methods.
+    - Resolving such disagreements requires examination of the HTO data, which can not be stored in the same `AnnData` object. 
+    Therefore, we do not currently provide any multiplexed libraries as `AnnData` objects.
```
jashapiro, line 69: If we want to say something explicit?
```suggestion
    Therefore, we do not currently provide any multiplexed libraries as `AnnData` objects.
    - In addition, the multiplexed data are not compliant with the standards for `CELLxGENE`(https://cellxgene.cziscience.com), which we have tried to match as closely as possible.
```

allyhawkins:
@jashapiro I accepted your suggestion, with some slight rewording.

jashapiro (approved):
LGTM!

PR #301, opened 2024-04-24
Prepare docs release for H5 extension change

allyhawkins:
## Purpose

This includes all docs changes related to the upcoming change to use `.h5ad` extension on the Portal. 
It also includes the citation changes and a new FAQ for why some samples don't have AnnData. 

### Issue addressed

#293 

## Checklist for publishing to `main`

- [x] This branch contains all commits/content ready to be released.
- [x] The most recent CHANGELOG entry includes a date.
If this release required a CHANGELOG entry, it should be today's date.

PR #303, opened 2024-04-25
Add backticks

sjspielman:
This is the one thing I caught that needed fixing in #301

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,7 +14,7 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 
 ## 2024.04.26
 
-* AnnData files now have the extension `.h5ad` instead of `.hdf5`.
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 
 ## 2024.04.26
 
-* AnnData files now have the extension `.h5ad` instead of `.hdf5`.
+* `AnnData` files now have the extension `.h5ad` instead of `.hdf5`.
```
allyhawkins, line 17:
```suggestion
* `AnnData` objects are stored in files that now have the extension `.h5ad` instead of `.hdf5`.
```

allyhawkins (approved):
I made an edit that I think makes it more explicit.

sjspielman:
I agree with the change!

PR #307, opened 2024-05-07
Update FAQ for samples with missing cell type reports

allyhawkins:
Since we are removing any libraries that have < 2 cells, the second bullet point in this FAQ no longer applies. Even if there's between 2-30 cells, the report and annotations will be present for SingleR and then CellAssign will say "Not run". We indicate this in the SCE contents section so I don't think we need those specifics here. 

I updated this to just mention when cell type annotations aren't present at all which would be cell lines.


sjspielman (approved):
> We indicate this in the SCE contents section so I don't think we need those specifics here.

I think if we want to communicate this further, we would want a different FAQ along the lines of "which cells in a library have annotations." We can add this in the future if questions come up about it, so I think your changes to this section are fine.

PR #322, opened 2024-07-08
Fix Changelog link

jashapiro:
Fixing a broken link (formatting) in the changelog.

PR #320, opened 2024-07-08
Prep docs for release of portal wide metadata

allyhawkins:
Closes #318 

Here I uncommented out the section of the docs that described the portal wide metadata and added a changelog entry. I kept the changelog entry pretty brief, but let me know if I should expand any more there? 

Before we actually release these docs we will need to add a date to the changelog.

PR #323, opened 2024-07-08
Fix broken links in changelog in development

allyhawkins:
Just mirroring the changes made in #322 and also fixing the link in the new changelog additions from #320.


sjspielman (approved):
nice, good catch!

PR #335, opened 2024-07-24
Fill in some docs gaps

### LINKED ISSUE #333: Add "Unclassified cell" to documentation
A question has come up about cell type annotations in OpenScPCA, and it turns out we are missing some docs here! We use "Unclassified cell" labels for cells that do not get run through cell typing, but this is not documented anywhere. We should note this both in the cell type annotation section and in the SCE contents section.

Relevant code is in this `scpca-nf` script: https://github.com/AlexsLemonade/scpca-nf/blob/main/bin/add_celltypes_to_sce.R

### LINKED ISSUE #334: Explain minimum gene cutoff in documentation
It seems that we do not fully explain our filtering procedure if `miQC` fails in the docs. We only state that the filtering method will be `Minimum_gene_cutoff` in the SCE contents table (https://scpca.readthedocs.io/en/stable/download_files.html?highlight=Minimum_gene_cutoff#metadata), but we don't explain this in processing. It would probably be worth adding a little more context here.

sjspielman:
Closes #333 
Closes #334 

This PR makes some small additions to expand documentation on cell type annotation and filtering as noted in the linked issues. I tried to keep changes to a minimum while still getting the relevant points across - did I succeed??


allyhawkins (approved):
LGTM

PR #331, opened 2024-07-23
Add `age_timing` to docs

allyhawkins:
Closes #326 

This PR adjust the description of the `age` column and adds the new `age_timing` column that we now will have in the sample metadata. I believe I adjusted all areas where we have the sample metadata table as mentioned in #326.

PR #336, opened 2024-07-25
Update age timing description to note that it is always diagnosis at diagnosis

### LINKED ISSUE #332: Update age_timing to note when diagnosis is used
Yea, I think that makes sense. We might want to add something in the user-facing docs about our assumption in case anyone is filtering on this column.

_Originally posted by @jaclyn-taroni in https://github.com/AlexsLemonade/ScPCA-admin/issues/933#issuecomment-2248006540_

`age_timing` will always be `diagnosis` if `disease_timing` is `Initial diagnosis`. This should be noted in the docs description.

allyhawkins:
Closes #332 

This updates the existing `age_timing` description added in #331 to clarify that this value will always be `diagnosis` if the sample disease timing is diagnosis. Is the way I described it clear?


sjspielman (approved):
LGTM, I think this is clear!

Issue #347, opened 2024-09-12
Update alevin-fry's references


DongzeHE:
Dear scPCA team,

Thank you very much for brining scPCA to the community and choosing alevin-fry as the preprocessing tool! 

I saw that the reference to alevin-fry might be outdated. It currently points to the preprint [He et al. 2021] (https://doi.org/10.1101/2021.06.29.450377) instead of our peer-reviewed paper [He et al. 2023](https://doi.org/10.1038/s41592-022-01408-3). It will be great if these references could be updated. Please let me know if you want me to submit a PR with the updated links! 

Best,
Dongze

jaclyn-taroni:
Hi @DongzeHE, thanks for bringing this to our attention! We would welcome a pull request to update these references.

PR #348, opened 2024-09-15
update alevin-fry and salmon references

DongzeHE:
Dear ScPCA team,

As discussed in #347, I updated alevin-fry references in the doc. Moreover, I noticed the doc says "we used `salmon alevin` and `alevin-fry`". I guess what we wanted to say was the `salmon` mapper, instead of `salmon alevin`, which is for quantification instead of mapping. 

Hope what I changed make sense. Thank you very much!

Best,
Dongze

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -3,7 +3,7 @@
 ## Why did we use Alevin-fry for processing?
 
 We aimed to process all of the data in the portal such that it is comparable to widely used pipelines, namely Cell Ranger from 10x Genomics.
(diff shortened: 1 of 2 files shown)


jashapiro (approved):
Hi Dongze, thank you for these updates! 

This looks good, but I did want to just comment about the salmon/salmon alevin changes. I believe the reason that we had used `salmon alevin` in the docs is in part because that command in salmon is the one we did use, in particular because it handles the cell barcodes, etc., but you are correct that we do not use the quantification portion of the software. I do think the change to just `salmon` is fine though, partly as it seems to make sense to link the top-level software (and we do use the index building portion as well as alevin), and also because the link change makes it more robust to any possible future changes in the documentation. The `alevin-fry` documentation is much more relevant to the options we are using at the moment anyway.

PR #349, opened 2024-09-16
update spellcheck action

jashapiro:
Apparently our old spellcheck action had some deprecated components that are now not allowed by github. Seemed a good time to switch this repo over to our internal action for this purpose.


.github/workflows/spell-check.yml
```diff
@@ -14,29 +13,32 @@ jobs:
   # This workflow contains a single job called "spell check"
   spell-check:
     runs-on: ubuntu-latest
-    container:
-      image: rocker/tidyverse:4.3.2
```
sjspielman, line 16: Do you want to add...

```suggestion
concurrency:
  # only one run per branch at a time
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

```

sjspielman (approved):
👍 
We can also remove the `scripts` directory at this point, since it only contains the spell check script.

.github/workflows/spell-check.yml
jashapiro: The spellcheck is fast enough that I don't expect this to be an issue. Getting notifications of cancelled jobs is more annoying than the extra seconds of processing, I think.

PR #350, opened 2024-09-17
Update CHANGELOG.md heading

jashapiro:
Add date of metadata changes.


jashapiro:
In this case we can do a direct PR from development to main, so I was planning that, but I can go the other way if desired!

jaclyn-taroni:
I don't think we want to put in the date yet. I think we want to file a PR from `development` -> `main` with the placeholder in it, and then update the date when it hits `prod` and we're ready to merge.

PR #363, opened 2024-10-31
Update changelog with counts and capitals

### LINKED ISSUE #362: Update changelog for coming reprocessing
In https://github.com/AlexsLemonade/ScPCA-admin/issues/986 we will be reprocessing all data with integer counts, so we need to add that to the changelog.

In addition, in https://github.com/AlexsLemonade/ScPCA-admin/pull/985, I am updating/normalizing some of the column names in the metadata (making capitalization more uniform), which we will want to note in the changelog as well

jashapiro:
Closes #362 

Here I added two bullet points to the changelog to cover the change to the count matrix and the column name changes that were made. I don't know if I love my wording for any of these, so feel free to tell me it is bad.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,13 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 <!-- PUT THE NEW CHANGELOG ENTRY RIGHT BELOW THIS -->
 <!-------------------------------------------------->
 
+## Upcoming
+
+* The raw count matrix is again rounded to integer values, in agreement with the documentation
+* Some project-specific metadata columns were renamed for uniformity
+  * columns previously labelled as `mycn_status` or similar with a gene name prefix are capitalized `MYCN_status`
+  * columns that were previously labeled with sentence case are now all lower case.  
+
 ## 2024.10.14
 
 * Some cell count-related fields in downloadable metadata have changed.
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 <!-- PUT THE NEW CHANGELOG ENTRY RIGHT BELOW THIS -->
 <!-------------------------------------------------->
 
+## Upcoming
+
+* The raw count matrix is again rounded to integer values, in agreement with the documentation
+* Some project-specific metadata columns were renamed for uniformity
+  * columns previously labeled as `mycn_status` or similar with a gene name prefix are capitalized `MYCN_status`
+  * columns that were previously labeled with sentence case are now all lower case.
```
jaclyn-taroni, line 20:
```suggestion
* Some project-specific metadata columns were renamed for uniformity:
  * Columns previously labeled as `mycn_status` or similar with a gene name prefix are capitalized `MYCN_status`.
  * Columns that were previously labeled with sentence case are now all lower case.  
```

docs/CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 <!-- PUT THE NEW CHANGELOG ENTRY RIGHT BELOW THIS -->
 <!-------------------------------------------------->
 
+## Upcoming
+
+* The raw count matrix is again rounded to integer values, in agreement with the documentation
```
jaclyn-taroni, line 17: If I'm right that this only impacted sc/snRNA-seq and if using code styling for those words is consistent with what we do elsewhere (I'm pretty sure it is):

```suggestion
* Recent versions of the raw count matrices in `SingleCellExperiment` and `AnnData` objects were not rounded. 
The raw counts are now rounded to integer values in agreement with the documentation.
```

jaclyn-taroni (approved):
I'd like to see a few small changes before this goes in, but I don't need to see it again.

Issue #367, opened 2024-11-14
Update metadata file description for multiplexed data


sjspielman:
We recently added a column `demux_cell_count_estimate` to the multiplexed metadata. Unfortunately, this made it into the docs as `demux_sample_cell_estimate`:

- https://github.com/AlexsLemonade/scpca-docs/blob/2ea40c44eb34a53d297edde57bc25d7eb9a8718d/docs/CHANGELOG.md?plain=1#L27
- https://github.com/AlexsLemonade/scpca-docs/blob/2ea40c44eb34a53d297edde57bc25d7eb9a8718d/docs/download_files.md?plain=1#L185

This issue tracks correcting this column name in the docs.

jaclyn-taroni:
It's not the docs that are wrong: https://data-lab-knowledge.slab.com/posts/2024-10-01-sample-count-discussion-5o843itw

sjspielman:
> It's not the docs that are wrong:

Well, now I don't understand this discrepancy. Is it even a discrepancy? 
Context: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/880#discussion_r1842759671

sjspielman:
Ah, is the implication `scpca-nf` is recording this wrong then?

jaclyn-taroni:
I think the download is not what was discussed as recorded in that Slab document (or at least as I understood it).

jashapiro:
Yes, it appears that the field name in slab was not used in the portal code. At this point, it seems the easiest change to make is the docs, especially given that the current field name isn't particularly _bad_.

jaclyn-taroni:
The other question would be if the calculation is correct. If it is correct, and it's just the name, we can just update the docs.

jashapiro:
> The other question would be if the calculation is correct. If it is correct, and it's just the name, we can just update the docs.

The calculation does appear to be correct. At the moment, there are some small (1-2 cell) discrepancies between what is on the portal and what was is in the input bucket, but I expect these are due to the changes associated with rounding. (The portal does not seem to yet have the updated SCPCP000009 files, confirmed by downloading a sample and seeing non-integer values)

jaclyn-taroni:
I propose we do the following:

- Update the relevant CHANGELOG entry and the sentence Stephanie highlighted via PR to `development`
- `development` -> `main`
- Cut a new release

[...]

PR #370, opened 2024-11-15
Replace `demux_sample_cell_estimate` with `demux_cell_count_estimate`

jaclyn-taroni:
Addresses #367 

My plan – as described in https://github.com/AlexsLemonade/scpca-docs/issues/367#issuecomment-2478895224 – is to merge `development` -> `main` after this and cut a new release. I don't believe a CHANGELOG entry is required, as this is correcting documentation about the current downloadable files.


jashapiro (approved):
LGTM. Thanks for handling this. I agree that we do not need a new CHANGELOG entry.

PR #373, opened 2024-11-20
Update bulk names for download illustrations

dvenprasad:
Closes #369 

Updated the bulk file names
Bumped up the quality of the images too

Issue #368, opened 2024-11-14
Change in bulk file names


jashapiro:
When looking into #367 and trying to understand missing events in https://github.com/AlexsLemonade/OpenScPCA-nf/pull/102, I saw that the names for the bulk files have changed. Where they were before named `bulk_metadata.tsv` and `bulk_quant.tsv`, they are now prefixed with the project name:  e.g., `SCPCP000001_bulk_metadata.tsv`, etc.

This change is not reflected in the docs, so we should update the docs. (I like the change, but was not aware of it.)

https://github.com/AlexsLemonade/scpca-docs/blob/2ea40c44eb34a53d297edde57bc25d7eb9a8718d/docs/download_files.md?plain=1#L29-L31

jaclyn-taroni:
I bet that the same update as described here is responsible: https://github.com/AlexsLemonade/scpca-docs/blob/main/docs/CHANGELOG.md#20240813

We should address what you've identified here and perhaps make the CHANGELOG entry more general (or at least give the bulk example, too).

jaclyn-taroni:
We need to change the images, too, unfortunately.

jaclyn-taroni:
@davidsmejia - do you know when this changed? I'm trying to understand if we need to correct an existing CHANGELOG entry or add a new one.

davidsmejia:
> @davidsmejia - do you know when this changed? I'm trying to understand if we need to correct an existing CHANGELOG entry or add a new one.

This change was deployed on 2024-07-31. This would have been part of the changes that were updated / documented in 2024-08-13. But the original change landed in the previous portal release / docs change.

PR #374, opened 2024-11-21
Update downloadable files section and 2024.08.13 CHANGELOG entry to reflect bulk filenames

jaclyn-taroni:
Stacked on #373 

Here, I'm updating the three instances where we discuss bulk files in the downloadable files section. I tried to be consistent with how we discuss accession identifiers in filenames in that section.

I am also updating the 2024.08.13 CHANGELOG entry, as this change went into effect on 2024-07-31 per https://github.com/AlexsLemonade/scpca-docs/issues/368#issuecomment-2479528007.


jashapiro (approved):
LGTM!

PR #434, opened 2025-07-17
Changelog entry about merged objects bug fix

### LINKED ISSUE #432: Add a changelog to note fixing the bug in the `cell_id` column for merged objects
We are updating all merged objects on the Portal to have the correct contents of the `cell_id` column. See https://github.com/AlexsLemonade/scpcaTools/pull/299. 

We will need to create a changelog entry to note the contents are now correct and do a new docs release. Note that the `development` branch currently has all the cart docs, so we should add this in a feature branch and then merge that into both `main` and `development`.

allyhawkins:
Closes #432 

Here I'm adding in the changelog entry needed to account for the bug fix in the merged objects. I will keep this in a draft state until we are ready to update the objects on the Portal and then I can update the date. 

Once this goes in, we will need to merge this branch (with the correct date) into `development`.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,11 @@ For more information about `AlexsLemonade/scpca-nf` versions, please see [the re
 <!-- PUT THE NEW CHANGELOG ENTRY RIGHT BELOW THIS -->
 <!-------------------------------------------------->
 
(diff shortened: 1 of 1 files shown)


sjspielman (approved):
LGTM - I trust that you can fill in the correct date when the time comes but feel free to tag me back if you want :)

PR #442, opened 2025-08-12
Turn on spell check

sjspielman:
In this PR I'm adding `the-manuscript` to the branches where the spell check action should be run. I'm pretty sure this will work to send this PR directly to `the-manuscript` branch (as in I don't think it needs to be committed to in `main`), but in case I'm wrong we'll find out soon enough as PRs come in :) 
I put it in quotes too, since don't know about that dash.


allyhawkins (approved):
🐈

PR #453, opened 2025-09-24
Update "uns" docs to explain excluded fields

sjspielman:
This PR is a companion PR to https://github.com/AlexsLemonade/scpcaTools/pull/322 (hence my choice of reviewer!).

It turns out that the current docs didn't actually mention yet anything about fields that are missing from anndata `.uns` slot but present in SCE metadata. So, I got to add it in! I rephrased a smidge to include a bullet point that says the `uns` is missing any metadata that couldn't be converted to python, but I don't actually state the name of those fields. I did, however, update the metadata table to indicate that it's a `DataFrame` to help traceability. 

Let me know if you think the level of detail here should shift at all!


jashapiro (approved):
LGTM, with one small wording update.

PR #465, opened 2025-11-24
Add new spaceranger images 

### LINKED ISSUE #461: Document Spatial Spaceranger as data format
We are going to have a new data format, `Spatial Spaceranger` when cart is deployed. We will need to ensure the docs account for this option. My initial thought is that this will just involve updating the download options section to indicate the new format. https://github.com/AlexsLemonade/scpca-docs/blob/development/docs/download_options.md#data-format

allyhawkins:
Closes #461 

Here I'm adding in the new images that replace `spatial` in the zip file names with `spaceranger` and updated some wording to reflect the new `Spaceranger` data format that is only an option when downloading spatial data. 

I updated the `Data format` section to reflect this change and then went through the `Download files` section everywhere we talk about "data format" to ensure any mentions of spatial are accurate.

### DIFF
--- a/docs/download_options.md
+++ b/docs/download_options.md
@@ -15,7 +15,7 @@ Only {ref}`one data format is currently supported for a single download<faq:Why 
 To obtain data in both `SingleCellExperiment` and `AnnData` formats, you will need to download these file formats separately.
 
 The only exception to this rule is when downloading spatial transcriptomics data, as the only data format available will be `Spaceranger`. 
-Spatial data can be coupled with single-cell data in either the `SingleCellExperiment` or `AnnData` formats, but not both.
+Spatial data can be coupled with single-cell data in either the `SingleCellExperiment` or `AnnData` format, but not with both.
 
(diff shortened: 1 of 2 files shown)


docs/download_options.md
```diff
@@ -14,6 +14,9 @@ You can learn more about using these object types from our FAQ sections on {ref}
 Only {ref}`one data format is currently supported for a single download<faq:Why can't I change the data format in My Dataset?>`, including when {ref}`downloading custom datasets<download_files:Custom datasets>`.
 To obtain data in both `SingleCellExperiment` and `AnnData` formats, you will need to download these file formats separately.
 
+The only exception to this rule is when downloading spatial transcriptomics data, as the only data format available will be `Spaceranger`. 
+Spatial data can be coupled with single-cell data in either the `SingleCellExperiment` or `AnnData` formats, but not both.
```
sjspielman:
```suggestion
You can download spatial data at the same time as single-cell data in either the `SingleCellExperiment` or `AnnData` formats, but not both.
```

PR #493, opened 2026-03-18
Restore FAQ for downloading programmatically

### LINKED ISSUE #492: Answer to "Can I download data from the Portal programmatically?" is missing
Presumably this should link to the R package?

allyhawkins:
Closes #492 

It looks like the text was present, but the FAQ title was in the wrong place. I'm filing this to `main` since I think we should do a small release with this change to fix the docs currently displayed on the Portal.


sjspielman (approved):
great catch!

Issue #496, opened 2026-04-10
Change title of FAQ:What does the Copy Link button do?


dvenprasad:
Due to changes outlined in https://github.com/AlexsLemonade/scpca-portal/issues/1798,  the current title of [this FAQ ](https://scpca.readthedocs.io/en/stable/faq.html#what-does-the-copy-download-link-button-do) doesn't make sense. 

Rename it to `How can I use the command line tools link to download data?`

allyhawkins:
Closed by #499

PR #499, opened 2026-04-15
Change Copy Download Link FAQ to reflect new UI changes

dvenprasad:
Addresses #496 

In addition to changing the FAQ title, I also made a minor tweak to the language so it's more in line with what users see on the portal.


allyhawkins (approved):
LGTM

PR #508, opened 2026-06-23
Add information about the filename differences for GEM-X flex data

### LINKED ISSUE #507: Add a section to download files for 10x flex multiplexed file structure
The file names are slightly different for any 10x flex libraries that are also multiplexed. Filenames for all data files from libraries that are 10x flex multiplexed will be <library_id>-<sample_id>_unfiltered.rds rather than `<library_id>_unfiltered.rds`. 

We should add some notes about this in the download files section that mentions the change in file name for these libraries. Similar to how we have a note on file-name exceptions for CITE-seq.

allyhawkins:
Closes #507 

For any libraries processed with GEM-X Flex, the individual files are going to have both the library id and the sample id as a prefix. The structure of the download is not any different, so we just need to note in the download files section that these file names will have both ids as opposed to just the library id. 

To do this, I added a line mentioning the different file name in every spot that we describe additional information or exceptions to the downloads (e.g., bulk or multiplexed). Let me know if you think we need more detail.

### DIFF
--- a/docs/download_files.md
+++ b/docs/download_files.md
@@ -10,6 +10,7 @@ Please see our {ref}`CHANGELOG <CHANGELOG:CHANGELOG>` for a summary of changes t
 Data can be downloaded by either downloading a [single project](#project-downloads), creating a [custom dataset](#custom-datasets), or by choosing one of the [Portal-wide download options](#portal-wide-downloads).
 For all data downloads, sample folders (indicated by the `SCPCS` prefix) contain the files for all libraries (`SCPCL` prefix) derived from that biological sample.
 Most samples only have one library that has been sequenced.
+If the data was generated using GEM-X Flex, filenames will contain a prefix with both the library and sample id (`SCPCL000000-SCPCS000000`). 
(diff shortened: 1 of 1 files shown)


docs/download_files.md
```diff
@@ -10,6 +10,7 @@ Please see our {ref}`CHANGELOG <CHANGELOG:CHANGELOG>` for a summary of changes t
 Data can be downloaded by either downloading a [single project](#project-downloads), creating a [custom dataset](#custom-datasets), or by choosing one of the [Portal-wide download options](#portal-wide-downloads).
 For all data downloads, sample folders (indicated by the `SCPCS` prefix) contain the files for all libraries (`SCPCL` prefix) derived from that biological sample.
 Most samples only have one library that has been sequenced.
+If the data was generated using GEM-X Flex, filenames will contain a prefix with both the library and sample id (`SCPCL000000-SCPCS000000`).
```
sjspielman: I would move this right below the bullet points line 26 or so), so the info is a bit more discoverable.

sjspielman (approved):
lgtm. Just one suggestion to move one of the sentences

